PR #3, opened 2019-07-16
use ref_id instead of lid in phrases module

kavitharaju:
The changes brought in the code expects the DB also to be changed accordingly. 
This might require code changes in inserting to  `_bible_cleaned` tables, in the source upload portion/module.


kavitharaju:
Also updated part where value is inserted to _bible_cleaned_ tables

PR #7, opened 2019-08-09
Autographamt api code

premkumaar:
Added previous and next chapter references when getting current chapter content


bijob89:
Accepted

PR #9, opened 2019-09-13
added call to get all  book and ch. and 2 issues

RevantCI:
Added a call to get all the books and chapters of the bible
Added a param to the get bibles call to get language based if query param language = true
Changed the code when returning the whole bible if usfm then also returning it in json format

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -1728,13 +1728,24 @@ def biblePattern(*argv):
     }
     return pattern
 
-def sortLang(lang,version):
+def sortLang1(lang,version):
     for i,item in enumerate(lang):
         if item['language'] == version["language"]["name"]:
             lang[i]["languageVersions"].append(version)
             break
     else:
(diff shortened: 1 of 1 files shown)


agmt/main.py
```diff
@@ -1728,7 +1728,7 @@ def biblePattern(*argv):
     }
     return pattern
 
-def sortLang(lang,version):
+def sortLang1(lang,version):
```
joelthe1: Can we have better names than `sortLang1` and `sortLang2`? You could add something more descriptive. Also, lets start using comments and Documentation Strings: https://www.python.org/dev/peps/pep-0008/#documentation-strings

RevantCI: Hi Joel,
Have renamed the functions to more descriptive ones and added Documentaion strings in all the functions added by me. Hope the commited code is fine. Let me know if there are any more changes required.
Revant

PR #13, opened 2019-10-04
Remove tokenization from upload source API

kavitharaju:
As the VO team uses the upload source API to add bibles to the common DB, it was taking extra time to do the tokenization every time a book is uploaded.
Now tokenization is done, when a translator chooses a book from the token translation page and the getTokensList is invoked for the first time.


agmt/main.py
```diff
@@ -906,10 +906,20 @@ def getTokenLists(sourceId, book):
     bookId = cursor.fetchone()[0]
     tablename = '_'.join(rst[0].split('_')[0:3]) + '_tokens'
```
RevantCI: Can remove this line, doesn't seem to be doing anything

Issue #48, opened 2020-03-12
Audio Bible Metadata api


RevantCI:
Get the List of audio bibles and the metadata associated with them 

**1.Fetch the list of audio bibles, can filter by language** 

_Url - /v1/ audiobibles?language=<language_code>_ 

 Returns - 
```
[{ 
      language: {name:<name>,code:<code>,id:<id>}, 
      audioBibles: 
         [{ 
              sourceId: <source_id>,  
              name:<name>, 
              url:<folder_url>, 
              format:<mp3>, 
              books: { gen:50, exo:40,...//Books ,# of chapters for this bible} 
              language:{name:<language_name>,code:<language_code>,id:<language_id>}, 
          }, //other audio bible metadata 
         ] 
     }, ...//Other languages 
 ]
```

**2. Add the audio bible metadata to the /bibles call**  
_Url - /v1/ bibles_ 
Add to each source/version the corresponding audio bible 
```
audioBible:
      { 
            name:<name>, 
            url:<folder_url>, 
            format:<mp3>, 
            books:{  gen:50, exo:40,...//Books for this bible} 
      } 
```
- The audio bible are split by chapter and are stored in a file server.
- Get the audio bible for each chapter by combining the url with book ,chapter and format

eg. https://raw.githubusercontent.com/Bridgeconn/vachancontentrepository/master/audio_bibles/hin/irv/1co/1.mp3

i.e. `<url>+<book>+"/"+<chapterNo>+"."+<format>`

**Table structure and data** 

CREATE TABLE audio_bibles ( 
    id bigint NOT NULL PRIMARY KEY, 
    source_id bigint NOT NULL, 
    name text NOT NULL, 
    url text NOT NULL, 
    books jsonb, 
    format text NOT NULL, 
    status bool NOT NULL DEFAULT true, 
    FOREIGN KEY (source_id) REFERENCES sources (source_id) 
); 

Sample data:
insert into audio_bibles values(1,78,'Hindi IRV Audio Bible', 'https://vachan.sgp1.digitaloceanspaces.com/audio_bibles/hin/irv/','{ 
        "mat":28, 
        "mrk":16, 
        "luk":24, 
        "jhn":21, 
        "act":28, 
        "rom":16, 
        "1co":16, 
        "2co":13, 
        "gal":6, 
        "eph":6, 
        "php":4, 
        "col":4, 
        "1th":5, 
        "2th":3, 
        "1ti":6, 
        "2ti":4, 
[...]

PR #49, opened 2020-03-12
Audio Bible Metadata api #48, sort dictionary words

RevantCI:
Get audio bible metadata from database as per specification in issue #48 
Sort dictionary word

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -2114,7 +2114,8 @@ def getSources():
 def biblePattern(*argv):
     try:
         languageName, languageCode, languageId, script, scriptDirection, localScriptName, sourceId, \
-            versionCode, versionName, version, metadata, updatedDate,status = argv
+            versionCode, versionName, version, metadata, updatedDate,  status, audio_name, audio_url, \
+				audio_format, audio_books = argv
     except Exception as ex:
         print(ex)
     pattern = {
@@ -2132,7 +2133,7 @@ def biblePattern(*argv):
             "longName": version
         },
         "metadata":metadata,
-        "audioBible":[],
+        "audioBible":{"name":audio_name,"url":audio_url,"format":audio_format,"books":audio_books},
         "updatedDate": updatedDate,
         "sourceId": sourceId,
         "status": status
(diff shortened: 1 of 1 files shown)


agmt/main.py
```diff
@@ -2868,5 +2876,120 @@ def getInfographics(languageCode):
...
+    for item in languageObject["books"]:
+        book = item.pop("book")
+        if book in bookObject.keys():
+            bookObject[book].append(item)
+        else:
+            bookObject[book]= [item]
+    languageObject["books"] = bookObject
+    return languageObject
+
+@app.route("/v1/videos", methods=["GET"])
```
kavitharaju: @RevantCI I see you have added a separate table in DB for bible vedios. Are you adding the details of this table in the sources table?
Also make sure you update the staging DB with the new table so that the code should work properly once it is deployed on staging.

RevantCI: Will update the staging DB with the new table. Am not modifying the sources table, the audio_bible table has a foreign key to sources table.

kavitharaju: I see, that audo files are directly linked to a bible in the sources table. 
I asked about videos. Are they similar(referencing a bible), if not, are they entered in sources table?

RevantCI: The videos are in a seperate table but reference language table by id so that the hindi videos are linked together etc. - https://github.com/Bridgeconn/vachan-api/issues/50

CREATE TABLE bible_videos ( 
    id SERIAL PRIMARY KEY, 
    books text NOT NULL, 
    url text NOT NULL, 
    title text NOT NULL, 
    description text NULL, 
    theme text NULL, 
    language_id bigint NOT NULL, 
   status bool NOT NULL DEFAULT true, 
    FOREIGN KEY (language_id) REFERENCES languages (language_id) 

);

kavitharaju: I feel that these videos should be treated as a content type. There should a content type for "videos" in the content_types table and just have one entry in the sources table, for this table. In the future, if we have different kind of vedio, or vedios from a different publisher, then we can have a separate table, which again will be entered in the sources table

RevantCI: ok,Can we discuss the schema you are proposing. Just slack me when you get some free time.

Issue #59, opened 2020-07-13
Upload translated token API


Joel-C-Johnson:
## User Story

As a user/admin/super admin, I need to able to upload a translated token file from an excel... So need an API to takes the JSON data from UI

json format sample

json [ {…}, {…} ]

[
0 : { "Token": "God ", "Translation": "esho", "Senses": "mashea"},
1 : {"Token": "Love", "Translation": "kathal", "Senses": "pranayam"}
]

kavitharaju:
As you have shown an array I assume that the correct structure of input JOSN will be, 
[
 { "Token": "God ", "Translation": "esho", "Senses": "mashea"},
 {"Token": "Love", "Translation": "kathal", "Senses": "pranayam"}
]

without the keys, 0 , 1 etc you showed in the example.
Is this correct?

Joel-C-Johnson:
yeah, it will be a list of objects...
eg :  [ {…}, {…} ]

kavitharaju:
new API url : `"/v1/autographamt/projects/bulktranslations"`
inputs: projectId, tokenTranslations

Sample input data 
```
"projectId":57,
"tokenTranslations":
[
	{"token": "അംഗങ്ങളോടോ",
	"translation":"to members",
	"senses": []},
	{"token": "അംഗങ്ങൾ",
	"translation":"members",
	"senses": []},
	{"token": "അംഗഹീനനായിട്ടോ",
	"translation":"without an organ",
	"senses": []}
]
}
```
Please note that senses should be an array of strings.

Joel-C-Johnson:
> new API url : `"/v1/autographamt/projects/bulktranslations"`
> inputs: projectId, tokenTranslations
> 
> Sample input data
> 
> ```
> {
> "projectId":57,
> "tokenTranslations":
> [
> 	{"token": "അംഗങ്ങളോടോ",
> 	"translation":"to members",
> 	"senses": ['to the members', 'to organs', 'is it to members']},
> 	{"token": "അംഗങ്ങൾ",
> 	"translation":"members",
> 	"senses": []},
> 	{"token": "അംഗഹീനനായിട്ടോ",
> 	"translation":"without an organ",
> 	"senses": []}
> ]
> }
> ```
> 
> Please note that senses should be an array of strings.

[...]

PR #62, opened 2020-07-15
Bulk token upload

kavitharaju:
new API as per https://github.com/Bridgeconn/vachan-api/issues/59

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,5 +61,6 @@ typings/
 .next
 
 agmt/__pycache__/*
-
+test/__pycache__/*
+vachan-venv/*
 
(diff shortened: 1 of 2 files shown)


kavitharaju:
Made changes as per @RevantCI  's below suggestions
```
Had a couple of suggestion and a doubt.
suggestion: You could add a doc string to the function
suggestion : as list if a python data type the UI person may not understand it, in javascript they call it json array
Line 759: can we change the message "Incorrect datatype. token-translations should be list" to "Incorrect datatype: tokenTranslations should be and json array"
2. This does not make sense as 'dbSense' is of type list with senses split in the previous line and 'senses' is of type string with multiple senses combined in line 768:
Line 791: if senses not in dbSenses:
Are you trying to not have duplicates then you can use something like this:
dbSenses = dbSenses + senses.split("|")
dbSenses = list(dict.fromkeys(dbSenses))```

kavitharaju:
Added check for empty string in senses list as per @Joel-C-Johnson 's requirement

Issue #64, opened 2020-08-05
Download tokens


Joel-C-Johnson:
## User Story

As a user/admin/super admin, I need to able to download the " tokens, translated tokens and senses "  in an excel for each book in an excel... So need an API that returns the JSON in below format 
[
 [token,translation, senses, concordance, translationNotes, translationWords], 
 [token,translation, senses, concordance, translationNotes, translationWords],
 [token,translation, senses, concordance, translationNotes, translationWords],
]

kavitharaju:
Wouldn't these information be enough to download as excel ?
```
[
[token,translation, senses],
[token,translation, senses],
[token,translation, senses],
]```

**Comment by Joel-C-Johnson** [comment]
> Wouldn't these information be enough to download as excel ?
> 
> ```
> [
> [token,translation, senses],
> [token,translation, senses],
> [token,translation, senses],
> ]```
> ```

yeah this format will help to download the files

**Comment by kavitharaju** [comment]
The API is implemented in the branch [bulk-DOWNLOAD-tokens](https://github.com/Bridgeconn/vachan-api/tree/bulk-DOWNLOAD-tokens).
The details are as follows:

GET method
URL: `/v1/tokentranslationlist/<projectId>/<book>` (somewhat similar to `/v1/tokenlist/<sourceId>/<book>` which fetches just the tokens of a book, without translations)
AUTH: Only accessible to a user who has been assigned that book in that project.
INPUT: projectId and 3 letter bookCode(lower or upper cases accepted)
SAMPLE OUTPUT:
```
[
 {'token': 'അകത്തേക്കു', 'translation': None, 'senses': None},
 {'token': 'അകത്ത്', 'translation': 'inside', 'senses': None},
 {'token': 'അകത്ത്', 'translation': 'inside', 'senses': ['into']},
 {'token': 'അകത്ത്', 'translation': 'inside', 'senses': None},
 {'token': 'അകത്ത് ചെന്ന്', 'translation': 'going inside', 'senses': None},
 {'token': 'അകന്നാണു', 'translation': None, 'senses': None},
 ...
]
```

Joel-C-Johnson:
as we are using an npm module(xlsx) for downloading the file in an excel it takes a list as input.
eg:
var array = 
[ 
  [token,translation, senses],
  [token,translation, senses]
]

var ws_data = array
var ws = XLSX.utils.aoa_to_sheet(ws_data);

[...]

PR #66, opened 2020-08-12
Bulk download tokens

kavitharaju:
Added new API `/v1/tokentranslationlist/<projectId>/<book>`.
#59 
#64

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -55,8 +55,8 @@ def get_db():                                                                   
 	current application context.
 	"""
 	if not hasattr(g, 'db'):
-		g.db = psycopg2.connect(dbname=postgres_database, user=postgres_user,
-			password=postgres_password,	host=postgres_host, port=postgres_port)
+		g.db = psycopg2.connect(dbname=postgres_database, user=postgres_user)
+			# password=postgres_password,	host=postgres_host, port=postgres_port)
 	return g.db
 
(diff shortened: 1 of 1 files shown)


agmt/main.py
```diff
@@ -1082,6 +1090,84 @@ def getTokenLists(sourceId, book):
...
+		assignments = cursor.fetchone()
+		if (not assignments) or (book.lower() not in assignments.split('|')):
+			return '{"success":false, "message":"UnAuthorized! You haven\'t been assigned this book/project"}'
+
+		cursor.execute("select s.table_name from sources as s join autographamt_projects as p \
+		on s.source_id = p.source_id where p.project_id=%s", (projectId,))
+		source_table = cursor.fetchone()[0]
+		tablename = source_table + '_tokens'
+
+		cursor.execute("select book_id from bible_books_look_up where book_code=%s", (book,))
```
RevantCI, line 1135: You can add a check to see if the book is valid by checking if its in the bible_books_look_up table, in case it is not you can return a error, you can do it before line 1127 where book is used. if it is not valid the query result will be empty

kavitharaju: Added

Joel-C-Johnson (approved):

RevantCI (approved):
The code seems fine.
It would be nice to have a document attached with the UI requriements giving details about the input json expected and expected return json. 
It will help with the review and it will help in the long term to keep a record as why the API was made.

PR #68, opened 2020-08-19
checked None for senses

udkumar:


agmt/main.py
```diff
@@ -714,7 +714,7 @@ def updateProjectTokenTranslations():
 			if senses == rst[2] and translation == rst[1]:
 				return '{"success":false, "message":"No New change. This data has already been saved"}'
 			dbSenses = []
-			if rst[2] != "":
+			if rst[2] != None or rst[2] != "":
```
kavitharaju, line 717: shouldn't the connective be AND ?

agmt/main.py
```diff
@@ -807,8 +808,10 @@ def bulkUpdateProjectTokenTranslations():
 							userId, senses))
 				else:
 					dbSenses = []
-					if rst[2] != "":
+					if rst[2] != None or rst[2] != "":
```
kavitharaju, line 811: shouldn't it be AND?

kavitharaju (commented):
Please confirm which should connective be used there? AND or OR?
I feel you need AND as both the condition should be satisfied.

kavitharaju (commented):
What about the check in `bulkUpdateProjectTokenTranslations()` line number `812` ?

kavitharaju (approved):

udkumar (commented):
We have tested with both but working with None only.... so updated the code

Joel-C-Johnson (approved):

PR #75, opened 2020-09-17
Write apis migration

### LINKED ISSUE #65: Write API's required for Migration
Make API's as required as per following documents
[Migration Document](https://bridgeconn.sharepoint.com/:w:/s/Vachan-API/Eb5GQEjTsDpInUbAZxHXrroBBgk82F6RIq8E1Jy07Qhjxg?e=vVDYlp)
[API Spedification Document](https://bridgeconn.sharepoint.com/:w:/s/Vachan-API/EQa5fqOOVFRLm0XZh-SHwKABc2jfcovjrZIclZjuyEDGZQ?e=NDC2kY)
**Current DB diagram** - [Interactive view](https://dbdiagram.io/d/5f325f46e1246d54aa2ce3cc)
Vachan DB

Testing can be done using these postman Collections
[Write APIs.postman_collection.zip](https://github.com/Bridgeconn/vachan-api/files/5245165/Write.APIs.postman_collection.zip)

RevantCI:
Write APIs 1-5 for the migration as per this [document](https://bridgeconn.sharepoint.com/sites/DevTeam-AgMT-VachanAPI/_layouts/15/Doc.aspx?OR=teams&action=edit&sourcedoc={B557C75A-BCF9-49A6-BC73-B0FC517DD2AF}) in teams
That is Commentary, Dictionary, Infographics, Audio , Videos

Fixed 2 small issues also
-get all bibles api audio bibles showing even if status false - /v1/bibles
-get commentary chapter api throwing error if no book introduction - /v1/commentaries/<sourceId>/<bookCode>/<chapterId>

Closes #65

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -2403,9 +2403,9 @@ def getBibles():
 	cursor = connection.cursor()
 	#use status param to filter by status, default only true
 	status = request.args.get('status')
-	query = "select s.source_id, v.revision, v.version_code, v.version_description,v.metadata, \
+	query = "select s.source_id, v.revision, v.version_code, v.version_description,s.metadata, \
 		l.language_id, l.language_name, l.language_code, local_script_name, script, script_direction, \
-			created_at_date, s.status, a.name, a.url, a.format, a.books from sources s left join \
+			created_at_date, s.status, a.name, a.url, a.format, a.books, a.status from sources s left join \
 				languages l on s.language_id=l.language_id left join versions v on s.version_id= \
 					v.version_id left join audio_bibles a on s.source_id=a.source_id where s.content_id=1 "
 	statusQuery = " and s.status=true"
@@ -2420,7 +2420,9 @@ def getBibles():
 	biblesList = []
 	language = request.args.get('language')
 	for s_id, ver, ver_code, ver_name, metadata, lang_id, lang_name, lang_code, loc_script_name, script, \
-		script_dir, updatedDate, status, audio_name, audio_url, audio_format, audio_books in cursor.fetchall():
+		script_dir, updatedDate, status, audio_name, audio_url, audio_format, audio_books, audio_status in cursor.fetchall():
+		if not audio_status:
+			audio_name = audio_url = audio_format = audio_books = None
 		biblesList.append(
 			biblePattern(
 				lang_name,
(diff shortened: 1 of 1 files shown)


agmt/main.py
```diff
@@ -2844,6 +2846,85 @@ def getBibleVerseText2(sourceId, verseId):
...
+		cursor.execute("insert into versions(version_code,version_description,revision) values(%s,%s,%s) returning version_id",
+				(abbreviation,name,revision,))
+		rst = cursor.fetchone()
+	return rst[0]
+
+def getLanguageId(cursor,language):
+	'''Get language id for given language code'''
+	cursor.execute("select language_id from languages where language_code=%s", (language,))
+	rst = cursor.fetchone()
+	return -1 if not rst else rst[0]
```
kavitharaju, line 2876: returning `None`, when language is not present would be more intutive that `-1`, I think.

agmt/main.py
```diff
@@ -2844,6 +2846,85 @@ def getBibleVerseText2(sourceId, verseId):
...
+			contentId = getContentId(cursor,'commentary')
+			version_id = getVersionId(cursor,abbreviation,name,revision)
+			#insert commentary into sources
+			cursor.execute('insert into sources (table_name, year, license, content_id, language_id, version_id,status) values \
+				(%s, %s, %s, %s, %s, %s,true)', (tableName, year, license, contentId, languageId,version_id,))
+			#insert commentary data into table
+			commentaryData = []
+			for row in commentary:
+				commentaryData.append((row['bookId'],row['chapter'],row['verse'],row['commentary']))
+			execute_values(cursor,sql.SQL('insert into {} (book_id, chapter, verse, commentary) values %s').format(sql.Identifier(tableName)),
```
kavitharaju, line 2916: use this sql.SQL() method to create all SQL statements for better data security

agmt/main.py
```diff
@@ -2910,7 +2991,7 @@ def getCommentaryChapter(sourceId,bookCode,chapterId):
 	try:
 		connection = get_db()
 		cursor = connection.cursor()
-		cursor.execute("select metadata->'Copyright' from sources s inner join versions v \
+		cursor.execute("select s.metadata->'Copyright' from sources s inner join versions v \
```
kavitharaju, line 2994: if you have the `sourceId` and if the only data you want is from the metadata column of the `sources` table, then why do you need the `JOIN` with versions table?

Also, the check for `content_id` can be avoided, as `source_id` is the Primary key and that identifies the exact row.

agmt/main.py
```diff
@@ -2844,6 +2846,85 @@ def getBibleVerseText2(sourceId, verseId):
...
+		name = req["name"]
+		abbreviation = req["abbreviation"]
+		revision = req["revision"]
+		license = req["license"]
+		year = req["year"]
+		language  = req["language"]
+		commentary = req["commentary"]
+		connection = get_db()
+		cursor = connection.cursor()
+		tableName = "%s_%s_%s_commentary" %(language.lower(), abbreviation.lower(), str(revision).replace('.', '_'))
```
kavitharaju, line 2893: its good to check for the presence of(or replace) space in any of the components of tablename as that will cause issue in DB.

For example, the user could enter "K J V" for abbreviation. 
Same issue was reported in source creation for Bible.

agmt/main.py
```diff
@@ -2957,6 +3040,57 @@ def getCommentaryChapter(sourceId,bookCode,chapterId):
...
+		name = req["name"]
+		abbreviation = req["abbreviation"]
+		revision = req["revision"]
+		license = req["license"]
+		year = req["year"]
+		language  = req["language"]
+		dictionary = req["dictionary"]
+		connection = get_db()
+		cursor = connection.cursor()
+		tableName = "%s_%s_%s_dictionary" %(language.lower(), abbreviation.lower(), str(revision).replace('.', '_'))
```
kavitharaju, line 3058: its good to check for the presence of(or replace) space in any of the components of tablename as that will cause issue in DB.

For example, the user could enter "K J V" for abbreviation. 
Same issue was reported in source creation for Bible.

agmt/main.py
```diff
@@ -3106,7 +3293,7 @@ def getInfographics(languageCode):
 		language_id = cursor.fetchone()
 		if not language_id or language_id is None:
 			return '{"success":false, "message":"Invalid language code"}'
-		cursor.execute("select table_name,metadata from sources s inner join versions v on \
+		cursor.execute("select table_name,s.metadata from sources s inner join versions v on \
```
kavitharaju, line 3296: I believe the `JOIN` with `versions` table can be avoided here

kavitharaju (commented):
- It would be good to use SQL query builder for building all the SQL statements for better security.
- I have only done code review. Shall do API testing on local system and update here

kavitharaju:
I see that you have used user authentication. It checks if the user is valid. But does not check the role of the user, like if it is `Super Admin` or not. Hence any of the registered user in the AgMT App would be able to access these write APIs. Was this done intentionally?

kavitharaju:
For the write API `localhost:8000/v1/sources/infographic`, after adding the first set of data, when I try to add a second set with new revision number I get error message `"Infographics already exists"`. From the code, I understand only onle set of infographics is expected for one language. Hope this is expected behaviour. Just pointing out to confirm

agmt/main.py
```diff
@@ -3133,7 +3320,41 @@ def getInfographics(languageCode):
...
+		req = request.get_json(True)
+		sourceId = req["sourceId"]
+		name = req["name"]
+		url = req["url"]
+		books = req["books"]
+		format = req["format"]
+		connection = get_db()
+		cursor = connection.cursor()
+		#Check if valid sourceId
+		cursor.execute("select table_name from sources where source_id=%s", (sourceId,))
```
kavitharaju, line 3337: you could add a check for `content_type` of source, to confirm it is `bible` as audios are supposed to be attached only with bible's sources

agmt/main.py
RevantCI: Removed the extra join, am using content_id to make sure they are trying to access a source of type commentary only using this api, will move the associated error message next to this query

kavitharaju:
@RevantCI 
> Hi Kavitha, for simple insert queries where the table or column name is not dynamic sql builder is not required as it does nothing without the format used to add the table or column name

Alright then. Need not change them

> If you have a simple way of testing it you can test it out.
> I found postman easy to use.

I tested the APIs locally using postman and added my review after that as comments above.

agmt/main.py
RevantCI: Have added a similar check for get commentary chapter

kavitharaju:
@RevantCI 
> Do you suggest we add a role check here, which one would you suggest we use for these write APIs?

I feel that adding auth without a rolecheck will not be very effective as the current check permits all users of AgMT. If you are checking for the role of the user, then I believe it should be the role `SUPER ADMIN`.  The other two roles we have at present, `AgMT Admin` and `AgMT Translator`,  are users who register via the AgMT App and hence should not be permitted write access to any content other than that specific to their organisations and projects. These contents are certainly not belonging to any particular user/org/project, but common to the whole Vachan. Hence, my suggestion `SUPER ADMIN`

samuelanthrayose (changes requested):
Did a basic code review. Looks good ! Only change I'd suggest is to have the Super Admin role check for the Write API's, as @kavitharaju has already suggested

samuelanthrayose:
@kavitharaju 
 
> On infographics API...
> 
> I am not suggesting that you change it. If the present behaviour is in line with the data we have now(or may get in future), then the current logic, of **being able to create only one infographic table for a language and being able to write to it only once**, is fine. In case we get more data, later, we may not be able to add it to DB via the current APIs and DB constraints. So it is your call, to keep it as it is or change. I just want to make sure this is the expected behaviour.

Yes, as per the present data and in the near future also, we might have only one set of infographics for a language. So the current behaviour should be fine.

kavitharaju (approved):

samuelanthrayose (approved):

PR #87, opened 2020-10-23
Feature error handling

kavitharaju:
Change Existing Error Handling with following effects
* Include Error response list for each API, in documentation
* Define DatabaseException, NotAvailableException, AlreadyExistsException and TypeException, so that status_code and Error names will be predefined and only error details needs to be set in different invokations.
* Add sample code in all functions showing how to raise each type of exceptions, to be used while API implementations
* Define custom exception handlers for RequestValidationError, and HttpExceptions so that error responses are of same JSON schema as the other Exceptions mentioned above

Add logger Configs


kavitharaju:
I see your point. We can make that change.

@joelthe1 Do you have an opinion on this?
Currently, if there is a GET request and there is no matching objects as per the request, to be returned, we return error with `404` status code saying `Requested Content Not Available`. 
@RevantCI is asking to change it to return an empty array(`[]`) with status code `200` or `204` (I am not in favor of `204`, though)

joelthe1:
What @RevantCI says makes sense. Let's keep 200 as the response code and an empty array `[]` in the body. The 404 is for endpoints that straight up are invalid.

kavitharaju:
I have removed all `404` exceptions from the APIs.
For all the `GET` methods we can return empty array `[]`, for not available content.
For all put methods, I am raising an `InCorrectDataException` within methods for not available contents. It will return a `422: Input Validation Error`, just like the ones thrown by the input validator of FastAPI. But here we can set the message in the `details` field of the exception.(Example: `{"error": "Input Validation Error", "details": "language code 'XYZ' not present"}`)

Is this fine?

RevantCI (approved):

PR #93, opened 2020-11-11
Use linter

kavitharaju:
- Adds pylint to App
- changed smaller files in app/ folder as per linting suggestions(not main.py. )
- Added git action to check linting (now the check fails because of main.py )


kavitharaju:
These changes are include in #94

PR #97, opened 2020-12-05
Versions APIs

kavitharaju:
- SQL scripts to create versions table in seed_DB creation script
- implementation of GET, PUT and POST methods of versions
- Test cases for Versions APIs
- Documentation of Change Log in versions wrt vachan-api-V1
- Uses fixtures in test module
- some small changes in the implementation of languages API
- Also an SQL script defining table structures related to AgMT suggestions module

### DIFF
--- a/app/crud.py
+++ b/app/crud.py
@@ -1,8 +1,7 @@
 ''' Place to define all Database CRUD operations'''
 import json
-from sqlalchemy import String
 from sqlalchemy.orm import Session
-# from  sqlalchemy.dialects.postgresql.JSON.Comparator import astext
+
 import db_models
 import schemas
 
--- a/app/crud.py
+++ b/app/crud.py
@@ -1,6 +1,8 @@
 ''' Place to define all Database CRUD operations'''
-
+import json
+from sqlalchemy import String
 from sqlalchemy.orm import Session
+# from  sqlalchemy.dialects.postgresql.JSON.Comparator import astext
 import db_models
 import schemas
 
@@ -12,7 +14,7 @@ def get_content_types(db_: Session, content_type: str =None, skip: int = 0, limi
     return db_.query(db_models.ContentType).offset(skip).limit(limit).all()
 
 def create_content_type(db_: Session, content: schemas.ContentTypeCreate):
-    '''Adds a row to table'''
+    '''Adds a row to content_types table'''
     db_content = db_models.ContentType(contentType = content.contentType)
     db_.add(db_content)
     db_.commit()
@@ -21,7 +23,7 @@ def create_content_type(db_: Session, content: schemas.ContentTypeCreate):
 
 def get_languages(db_: Session, language_code = None, language_name = None, #pylint: disable=too-many-arguments
     language_id = None, skip: int = 0, limit: int = 100):
-    '''Fetched all rows, with pagination'''
+    '''Fetches rows of language, with pagination and various filters'''
     if language_code and language_name:
         return db_.query(db_models.Language).filter(
             db_models.Language.code == language_code.lower(),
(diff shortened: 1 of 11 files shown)


kavitharaju:
The SQL script defining table structures related to AgMT suggestions module,  is not finalized and is being discussed. That is pushed so as to save it in github repo. Once the structure is finalized, it will be included in seed_DB creation script and the separate script will be removed from repo.

app/crud.py
```diff
@@ -21,21 +22,15 @@ def create_content_type(db_: Session, content: schemas.ContentTypeCreate):
...
-    '''Fetched all rows, with pagination'''
-    if language_code and language_name:
-        return db_.query(db_models.Language).filter(
-            db_models.Language.code == language_code.lower(),
-            db_models.Language.language == language_name.lower()).all()
-    if language_name:
-        return db_.query(db_models.Language).filter(
-            db_models.Language.language == language_name.lower()).all()
+    '''Fetches rows of language, with pagination and various filters'''
+    query = db_.query(db_models.Language)
```
joelthe1, line 26: Is it possible to have the filter operations chained here itself? Also, how does SQLAlchemy behave if we pass in `None` as value to filter with?

app/main.py
```diff
@@ -291,54 +294,90 @@ def edit_language(lang_obj: schemas.LanguageEdit = Body(...), db_: Session = Dep
...
-#       raise VachanApiException(name="Database Error", detail=str(e), status_code=502)
-#   return {"message" : f"Updated version field(s)", "data": None}
+@app.post('/v2/versions', response_model=schemas.VersionUpdateResponse,
+    responses={502: {"model": schemas.ErrorResponse}, \
+    422: {"model": schemas.ErrorResponse}, 409: {"model": schemas.ErrorResponse}},
+    status_code=201, tags=["Versions"])
+def add_version(version_obj : schemas.VersionCreate = Body(...),
+    db_: Session = Depends(get_db)):
+    ''' Creates a new version '''
+    log.info('In add_version')
```
joelthe1, line 332: Is there a reason you want this to be `log.info` as opposed to `log.debug`?

app/main.py
```diff
@@ -291,54 +294,90 @@ def edit_language(lang_obj: schemas.LanguageEdit = Body(...), db_: Session = Dep
...
+            version_obj.revision = 1
+        if len(crud.get_versions(db_, version_obj.versionAbbreviation,
+            revision =version_obj.revision)) > 0:
+            log.error('Error in add_version')
+            raise AlreadyExistsException("%s, %s already present"%(
+                version_obj.versionAbbreviation, version_obj.revision))
+        return {'message': "Version created successfully",
+        "data": crud.create_version(db_=db_, version=version_obj)}
+    except SQLAlchemyError as exe:
+        log.error('Error in add_version')
```
joelthe1, line 345: In all of these `log.error` statements we should also log the actual error message.

joelthe1: Also, in some of these cases it is more useful to use `logging.exception`

joelthe1: Also, what happens when you raise the exception like this? Does FastAPI catch that and return a 5xx error?

kavitharaju: The actual error message is being logged at the exception handler. 
I have changes these to `log.exception` to get the error message along with traceback in log. It is inlcuded in the other PR, [sources API ](https://github.com/Bridgeconn/vachan-api/pull/98).

I have added exception handlers to these exceptions, which re-formats all of these into a common response structure with ```{"error": <error_name>,
 "details": <details obtained from the actual error message or that composed within the App>}```
So that there is consistancy in how the client receives error messages.

app/main.py
kavitharaju: I have folllowed a pattern of using `log.info` to record the method name and `log.debug` to record the complete parameter list the method recieved at the start of every method. 
This way, if we just want to record all the actions going on at server, setting the log level to info will do.
If we also want to know the details of all parameters passed(this can get very long for data upload API and all), we could use logging level, debug.

app/crud.py
kavitharaju: `None` would be treated as `null`, I think, and it will look for matches with `null` in DB.

What do you mean by _chaining here itself_ ? Compose the query in one line?

The way I am doing it now, we draft the base query and add filters to it depending on the optional arguments we have and run the query, only once, at the end(in the return statment, when we add `.all()`, `.first()` etc to query)

joelthe1: Ah.. That makes sense. As long as we are executing a singly query, it looks good.

app/main.py
joelthe1: Makes sense.

app/main.py
joelthe1: So usually we would run the server at log level `info`, unless you were thinking otherwise.

kavitharaju: I was thinking of running server at `warning` level. But `info` would work too. It would be logging all the input params, but would record all the methods being called.

PR #98, opened 2020-12-08
Sources APIs

kavitharaju:
This PR includes
- Changes in versions-api PR #97 
- DB table creation script for sources
- implementation of GET, PUT and POST apis for sources
- testcases for above sources APIs
- small changes in versions tests
- documentation of changes brought in source wrt vachan-api V1
- updation in logging to log execption traceback also

### DIFF
--- a/app/crud.py
+++ b/app/crud.py
@@ -1,9 +1,11 @@
 ''' Place to define all Database CRUD operations'''
 import json
+import sqlalchemy
 from sqlalchemy.orm import Session
 
 import db_models
 import schemas
+from custom_exceptions import NotAvailableException
 
 def get_content_types(db_: Session, content_type: str =None, skip: int = 0, limit: int = 100):
     '''Fetches all content types, with pagination'''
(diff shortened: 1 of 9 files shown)


app/crud.py
```diff
@@ -86,7 +88,7 @@ def create_version(db_: Session, version: schemas.VersionCreate):
     return db_content
 
 def update_version(db_: Session, version: schemas.VersionEdit):
```
joelthe1, line 90: What happens if there is some unforeseen exception raised in this (or other functions)? Do you intend to capture errors at a higher level (i.e. at the level where these functions are called?

kavitharaju: Yes, all exceptions are caught at the main module these functions are called from. There we are catching un-forseen exceptions as `Exception` and raising a GenericException from them
```
    except Exception as exe:
        log.exception('Error in add_contents')
        raise GenericException(str(exe)) from exe
 ```
 This has a handler that converts it to the pattern 
```{ "error": "Error",
      "detail: str(exe)}```
  with status code `500` for Internal Server Error.

The other Errors we have defined are [here](https://github.com/kavitharaju/vachan-api/blob/sources-APIs/app/custom_exceptions.py)

PR #99, opened 2020-12-14
Bible book lookup

kavitharaju:
- table creation script and seed data for bible_books_look_up table
- implementation of GET method for the table (`/v2/lookup/bible/books`)
- tests for the above API
- documentation of changes brought in


db/bible_books.csv
```diff
@@ -0,0 +1,67 @@
...
+31,"obadiah","oba"
+32,"jonah","jon"
+33,"micah","mic"
+34,"nahum","nam"
+35,"habakkuk","hab"
+36,"zephaniah","zep"
+37,"haggai","hag"
+38,"zechariah","zec"
+39,"malachi","mal"
+40,"matthew","mat"
```
joelthe1, line 41: Not sure where we are using this list but we should be aware that ID `40` doesn't exist in the USFM spec (https://github.com/ubsicap/usfm/issues/99)

Ideally we can just borrow from here: https://github.com/ubsicap/usfm/blob/master/docs/identification/books.rst

db/seed_DB.sql
```diff
@@ -30,3 +30,28 @@ CREATE TABLE public.versions (
     metadata jsonb,
     UNIQUE(version_code, revision)
 );
+
+
+CREATE TABLE public.sources (
+    source_id SERIAL PRIMARY KEY,
+    table_name text UNIQUE,
+    year integer NOT NULL,
+    license text DEFAULT 'CC BY SA'::text,
```
joelthe1, line 39: Do we have another place where we store the full text of licenses?

db/bible_books.csv
kavitharaju: Oh, Okay. We can change that. We have been following this usage in V1. In V2, these Ids are planned to be used only internally in Database as Foreign Key. The APIs are supposed to use the Bible Name Codes. But We can certainly change this in case it is to be used at some point

db/seed_DB.sql
kavitharaju: No we don't. May be we should have a table to store licenses.

db/bible_books.csv
kavitharaju: Changed.

db/seed_DB.sql
joelthe1: I think we should have a separate table for licenses and allow users to add/modify there. And then use that to link to licenses in the sources.

joelthe1: I'll let you take note of this and do it as another PR.

kavitharaju: Sure.

PR #101, opened 2020-12-17
Commentary APIs

kavitharaju:
This PR includes
- logic to handle dynamic table creation and mapping via SQL alchemy
- implementation of GET, PUT and POST methods of commentaries
- test for the above
- documentation of changes brought in , w,r,t V1
- Make few changes in logger module to make it able to be used from all other modules
- excludes testing of sources API from git actions
   Because the sources API has been updated to raise error when any of the dynamic table creation and mapping are not implemented. They will be implemented only during the implementation of respective API sets(bibles, dictionaries etc). So disabling tests for it, until then.

### DIFF
--- a/.github/workflows/run-test.yml
+++ b/.github/workflows/run-test.yml
@@ -39,5 +39,5 @@ jobs:
           mkdir logs
           touch logs/API_logs.log
           cd app 
-          python -m pytest
+          python -m pytest--ignore=test/test_sources.py
 
--- a/.github/workflows/run-test.yml
+++ b/.github/workflows/run-test.yml
@@ -39,5 +39,5 @@ jobs:
           mkdir logs
           touch logs/API_logs.log
           cd app 
-          python -m pytest--ignore=test/test_sources.py
+          python -m pytest --ignore=test/test_sources.py
 
--- a/app/custom_exceptions.py
+++ b/app/custom_exceptions.py
@@ -6,6 +6,8 @@ class GenericException(Exception):
     def __init__(self, detail: str):
         super().__init__()
         self.name = "Error"
+        print("comes in GenericException")
+        print(detail)
         self.detail = detail
         self.status_code = 500
 
--- a/app/custom_exceptions.py
+++ b/app/custom_exceptions.py
@@ -6,8 +6,6 @@ class GenericException(Exception):
     def __init__(self, detail: str):
         super().__init__()
         self.name = "Error"
-        print("comes in GenericException")
-        print(detail)
         self.detail = detail
         self.status_code = 500
 
--- a/app/main.py
+++ b/app/main.py
@@ -608,7 +608,8 @@ def get_bible_book(book_id: int = None, book_code: schemas.BookCodePattern = Non
     responses={502: {"model": schemas.ErrorResponse},
     422: {"model": schemas.ErrorResponse}}, status_code=200, tags=["Commentaries"])
 def get_commentary(source_name: schemas.TableNamePattern, book_code: schemas.BookCodePattern = None, #pylint: disable=too-many-arguments
-    chapter: int = None, verse: int = None, last_verse: int = None,
+    chapter: int = Query(None, ge=-1), verse: int = Query(None, ge=-1),
+    last_verse: int = Query(None, ge=-1),
     skip: int = Query(0, ge=0), limit: int = Query(100, ge=0), db_: Session = Depends(get_db)):
     '''Fetches commentries under the specified source.
     Using the params bookCode, chapter, and verse the result set can be filtered as per need, like in the /v2/bibles/{sourceName}/verses API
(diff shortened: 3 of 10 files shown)


kavitharaju:
#### Issue in testing dynamic table creation

For testing, we normally use a separate database transaction, which is rolled back after each test function, there by, undoing all DB operations we performed during the test.

In the testing of APIs of dynamic tables(bibles, commentaries etc), the source creation APIs, called by the test client, create new tables in DB. These table creation operations are performed not as transactions attached to the open session, but are bound to the engine(that is how it is done in SQLAlchemy), which makes it not possible to undo by rolling back the session or transaction. 

The effect it makes is that, the tables thus created would remain in the DB even after testing. They will be empty as we are able to roll back all insert or update operations on it. We have created the table creation procedure such that, this wont throw an error when a new table has to be created in same name(when we run the same test next time). But it does show a warning like shown below when we create a new table of same name. 

```SAWarning: This declarative base already contains a class with the same class name and module name as db_models.hin_TTT_1_commentary, and will be replaced in the string-lookup table.```

If we are running tests on local machines, or on git actions, I think this can be overlooked. It will only be an issue if we are to run the tests on (staging) servers. Then also it wont create any issue that prevents the normal functioning of the app, just that a few empty tables will be left as residue in the staging database.

File: app/custom_exceptions.py [hunk 1]
```diff
@@ -6,6 +6,8 @@ class GenericException(Exception):
     def __init__(self, detail: str):
         super().__init__()
         self.name = "Error"
+        print("comes in GenericException")
```
joelthe1, line 9: What are these lines for?

app/main.py
```diff
@@ -14,24 +11,31 @@
...
+    '''logs and returns error details'''
+    log.error("Request URL:%s %s,  from : %s",
+        request.method ,request.url.path, request.client.host)
+    log.exception("%s: %s",'Error', str(exc))
+    if status_code in exc:
+        status_code=exc.status_code
+    else:
+        status_code = 500
+    return JSONResponse(
+        status_code =status_code,
```
joelthe1, line 36: Doesn't the linter also catch such cases where there is space missing before/after an `=` sign?

app/main.py
```diff
@@ -397,6 +405,9 @@ def add_source(source_obj : schemas.SourceCreate = Body(...),
     except NotAvailableException as exe:
         log.exception('Error in add_source')
         raise exe from exe
+    except GenericException as exe:
+        log.exception('Error in add_source')
```
joelthe1, line 409: Ideally you would alter the message to give some idea of the error. For e.g. 
```
except AlreadyExistsException as exe:
        log.exception('Error in add_source')
```
you could say something like `log.exception('The source with ID %s already exists' % id)`

.github/workflows/run-test.yml
```diff
@@ -39,5 +39,5 @@ jobs:
           mkdir logs
           touch logs/API_logs.log
           cd app 
-          python -m pytest
+          python -m pytest --ignore=test/test_sources.py
```
joelthe1, line 42: Is there a reason for ignoring this test?

kavitharaju: Yes. As I have told [here](https://github.com/Bridgeconn/vachan-api/pull/101#issue-541832307),
Excludes testing of sources API from git actions:
Because the sources API has been updated to raise error when any of the dynamic table creation and mapping are not implemented. They will be implemented only during the implementation of respective API sets(bibles, dictionaries etc). So disabling tests for it, until then.

The work flow in the App is that, when ever a new source is created, corresponding entires are made in sources table, and a new table of required schema is created dynamically in the Database under the `source_name`/`table_name` in the sources table row. 

Now we are working on defining schema and mapping these dynamic tables and crud operations on it. started with `commentaries` tables.

app/main.py
kavitharaju: Those kind of descriptions of the error are included in the `detail` part of the error when they are [raised](https://github.com/kavitharaju/vachan-api/blob/5de8f39a3f28c226eee66290b1450132a8ee828c/app/main.py#L173).
And are logged in the [exception handler](https://github.com/kavitharaju/vachan-api/blob/5de8f39a3f28c226eee66290b1450132a8ee828c/app/main.py#L75). 
Thought writting the log statement once in the handler is better than writting it in every place it occurs.

A sample log entry looks like this
```
12/18/2020 11:36:24 AM|main.py:315|ERROR   : Error in add_version
Traceback (most recent call last):
  File "/home/kavitha/Documents/vachan-api/app/main.py", line 308, in add_version
    version_obj.versionAbbreviation, version_obj.revision))
custom_exceptions.AlreadyExistsException
12/18/2020 11:36:24 AM|main.py:79|ERROR   : Request URL:POST /v2/versions,  from : testclient
12/18/2020 11:36:24 AM|main.py:80|ERROR   : Already Exists: XYZ, 1 already present
```

app/main.py
kavitharaju: No. Pylint doesn't seem to mind that. 
It only points out trailing spaces and lines.

app/custom_exceptions.py
kavitharaju: Included it while developing and debuging. Shall remove.

PR #103, opened 2020-12-22
Do not allow duplicate organization name

kavitharaju:
Issue was reported that users were able to create multiple organizations with same name.
Existing implementation
- checks if combination of Org name and Org email were unique
- did not remove spaces around org names and other values

Changes made
- strips(remove preceeding and trailing spaces) org names, emails and address in `createOrganisations()`
- ~~expects org names to be unique instead of org name, email combination~~

fixes
- #104 
- #105

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -392,7 +392,8 @@ def createOrganisations():
 
 		cursor.execute("select user_id from autographamt_users where email_id=%s", (email,))
 		userId = cursor.fetchone()[0]
-		cursor.execute("select status from autographamt_organisations where organisation_name=%s ", (organisationName, ))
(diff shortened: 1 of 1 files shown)


rinsingh:
What happens if the same org name is created for another email id? Is org name checked for uniqueness for an email id or across the system

kavitharaju:
> What happens if the same org name is created for another email id? Is org name checked for uniqueness for an email id or across the system

It wasn't before. But with this fix, it is required to be unique across the system

rinsingh:
To allow multiple orgs to be associated with an email id, check for uniqueness of org name for an email id. Org name + email id must be unique. 
Duplicate org names can be created with separate email id.

rinsingh (approved):

PR #116, opened 2021-02-01
V2 documentations

kavitharaju:
Improving API docs
- Checked and corrected all API doc strings
- Added response messages of all POST and PUT APIs to the API documentations
- Added example values for data objects returned by all APIs
- Added example values for all inputs(Path, Query and Body) in all APIs. These values are composed so that someone using the API docs to test the APIs can run all APIs with minimal efforts

Other Documentations
- Added change log for infographics, bible and license tables
- Updated ER diagram with all tables defined so far
- Added a Manual Testing Guideline

Code Changes
-  Changed content types `infographics` to `infographic` and `bible_video` to `biblevideo` to maintain uniformity with other content types. Made required changes in code and tests for these changes
- Format changes for references in return objects of bible/books and bible/verses 
- Remove `details` and `active` keys in dictionary words GET result when `words_only` flag is set to `TRUE`


app/crud.py
```diff
@@ -505,7 +505,7 @@ def get_infographics(db_:Session, source_name, book_code=None, title=None, #pyli
     '''Fetches rows of infographics from the table specified by source_name'''
     if source_name not in db_models.dynamicTables:
         raise NotAvailableException('%s not found in database.'%source_name)
-    if not source_name.endswith('infographics'):
+    if not source_name.endswith('infographic'):
```
joelthe1, line 508: For all these constant strings you use, it would be better if you use `enum` like [here](https://docs.python.org/3/library/enum.html#creating-an-enum).

You could create the `enum` elsewhere and import it here. So then in the future we can change things in a central place. What do you think?

kavitharaju: okay. Shall make that change

kavitharaju:
@joelthe1 

I have updated this PR with the changes for using ENUM in place of string literals like "bible", "commentary" etc

Also, broke down the crud moudle into multple files as you had asked

Changed one file
```
app/
- crud.py
```

to a folder as below
```
app/
- crud/
	- contents_crud.py
	- structurals_crud.py
	- npl_crud.py
```

**structurals_crud**: Place to define all Database CRUD operations for tables: Content_types, Languages, Licenses, versions, sources and bible_book_loopup

**contents_crud**: Place to define all Database CRUD operations for content tables: bible, commentary, infographic, biblevideo, dictionary etc

**nlp_crud**: CRUD operations with more text processing and linguistics. 

You had asked me to divide the crud module to one file per table. But I haven't done that, instead divided the tables into just 3 sets, 
- structuals, which contain the predefined tables in the DB structure
- contents, which contain the content carrying dynamically created tables
- nlp, which contain tables and operations related to AgMT stuff like tokenization, draft generartion, alignment, model building etc

if you think it would be better to have one file per table as you suggested, I can change this.

joelthe1:
I think the changes look good. But could you add the details you mentioned in your comment as comments on the top of the three new files you created.

Also, adding this here since this PR relates to documentation: I think we should add an `ARCHITECTURE.md` file for the `v2` vachan-api. You find more details from here:
- https://matklad.github.io//2021/02/06/ARCHITECTURE.md.html
- https://www.uzpg.me/tech/2020/07/21/architecture-md.html

I also, think we should add something like this for all our projects, especially USFM-Grammar.

joelthe1:
LGTM 👍

PR #117, opened 2021-02-02
Bug fix in AgMT User-Book Assignments

kavitharaju:
- changed response message to "Books assigned"
- updated value updation to add to existing books

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -627,28 +627,27 @@ def createAssignments():
 		cursor.execute("insert into autographamt_assignments (books, user_id, project_id) \
 			values (%s, %s, %s)", (books, userId, projectId))
 		connection.commit()
-		cursor.close()
 		return_message = '{"success":true, "message":"User added"}'
 
 	if action in ["assign", "unassign"] and not rst:
 		return '{"success":false, "message":"User not added to project yet"}'
-
-	old_books = rst[1].split('|')
-	if action == "assign":
-		updated_books = list(set(old_books+books))
-		books = [buk for buk in updated_books if buk != ""]
-		books = "|".join(books)
-		return_message = '{"success":true, "message":"Books assigned"}'
-	if action == "unassign":
-		for buk in books:
-			if buk not in old_books:
-				return '{"success":false, "message":"'+buk+' not in assigned list"}'
-			old_books.remove(buk)
-		books = "|".join(old_books)				
-		return_message = '{"success":true, "message":"Books unassigned"}'
-	cursor.execute("update autographamt_assignments set books=%s where user_id=%s and \
-		project_id=%s", (books, userId, projectId))
-	connection.commit()
+	if action in ["assign", "unassign"]:
+		old_books = rst[1].split('|')
+		if action == "assign":
+			updated_books = list(set(old_books+books))
+			books = [buk for buk in updated_books if buk != ""]
+			books = "|".join(books)
+			return_message = '{"success":true, "message":"Books assigned"}'
+		if action == "unassign":
+			for buk in books:
+				if buk not in old_books:
+					return '{"success":false, "message":"'+buk+' not in assigned list"}'
+				old_books.remove(buk)
+			books = "|".join(old_books)				
+			return_message = '{"success":true, "message":"Books unassigned"}'
+		cursor.execute("update autographamt_assignments set books=%s where user_id=%s and \
+			project_id=%s", (books, userId, projectId))
+		connection.commit()
 	# send email notification
 	try:
 		cursor.execute("SELECT first_name, email_id from autographamt_users where user_id=%s",(userId,))
@@ -680,7 +679,6 @@ def createAssignments():
 	except Exception as e:
 		print(e)
 		return '{"success":false, "message":'+str(e)+'}'
-
 	return return_message
 
 @app.route("/v1/autographamt/projects/assignments", methods=["DELETE"])
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -606,10 +606,13 @@ def createAssignments():
 	req = request.get_json(True)
 	userId = req["userId"]
 	projectId = req["projectId"]
-	books = req["books"]
-	# action = req["action"]
+	input_books = req["books"]
+	action = req["action"]
+	books = input_books
 	if not isinstance(books, list):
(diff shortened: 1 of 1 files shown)


Joel-C-Johnson:
I have tested this PR and found the following issue.
We are using this API in two function 
## 1. For adding a user in the project.
   API -  v1/autographamt/projects/assignments
  Header Request - {"projectId":"2","userId":4,"books":[]}
  Response - {"success":true, "message":"Books assigned"}

    In the UI there is a button for adding a user in the project
    1. Click on a button 
    2. it will list all the users in agmt.
    3. when selecting a user, the user will be added to that project
    
Here in the 1st function, we are trying to add a user to the project, so the message will be "the user added to the project" 

## 2. Assigning books to a user.
        API - /v1/autographamt/projects/assignments
     
        a. Here, when we are assigning books first time to a user it will work
            Header Request - {"projectId":"2","userId":4,"books":["2jn","1jn"]}
            Response -  {"success":true, "message":"Books assigned"}
            ----> Works fine as expected  

       b. Here, when we are removing the assigned book for a user ( uncheck 1jn from UI )
           Header Reuest - {"projectId":"2","userId":4,"books":["1jn"]}
           Response - {"success":true, "message":"Books assigned"}
           Here the books is not removed from the user.

kavitharaju:
@Joel-C-Johnson 
At the server side it is implemented to perform a single function, which is **to assign a set of books to a user in a project**.

I see that you are using the same function for 3 operations
1. adding a user to a project, by passing empty books list
2. assigning books to a user, by passing a set of books(the intended use)
3. un-assigning a set of books from a user by passing some other books

The way it was implemented before this fix, every time this API was invoked, it was re-writing the books list in the table against that user. So following worked fine
2. assigning books to a user
3. un-assigning a set of books from a user
but when the same user was added again to the project, passing an empty list(option 1, above), it used to over write the existing books list with the newly passed empty list, which was the cause for one of the problems you raised in your original issue, that books are being un-assigned.

Now after this change, I am not over writing the list of books, but adding the newly sent list to the existing one. So I hope, option 1, doesn't cause the problem you raised in your original issue. But option 3 miss functions now.

I believe we first need to clearly define what this API need to do. 
Also do we need separate APIs to assigning users to projects and un-assigning users from projects?

Joel-C-Johnson:
OK. as u said we are using this API for 3 operations. So it will be good to separate the API and we can return the message accordingly. Correct me if a wrong

I think the above API can able to assign the books, now for adding a user to a project and for removing the books from the user it will be good to have a separate api's..

kavitharaju:
In the code base I see this
```
@app.route("/v1/autographamt/projects/assignments", methods=["POST"])
def createAssignments():
	req = request.get_json(True)
	userId = req["userId"]
	projectId = req["projectId"]
	books = req["books"]
	# action = req["action"]
```
I think Bijo had plans to specify a required action also for this API, which was then commented out.
May be we can implement that.
How about we use `assign`, unassign` and `add_user` options for this `action` input and perform the 3 operations in the same API?

Joel-C-Johnson:
As we discussed, we are using the same API, passing a value in the API for the three functions, I will list some response messages that needs to come from the API side. 

action has two values-   "add_user" and "assign" 

URL - /v1/autographamt/projects/assignments
Header Request - {"projectId":"2","userId":3,"books":[],"action":"assign"}
The above is the API and the Header Request

1. If books list is empty and action is "assign" 
    - Do a check in the database if any books are assigned to this user, 
    a. if YES, -
     you can remove all the books from this user (This happens when we uncheck all the assigned books from the UI   resulting the books will be empty list "[ ]")   This operation will be treated as Unassigning books for the user.
      Message response: - "Assigned books updated" 
    b. if NO, -
     You can return a message saying "Please select any book" or not doing any further steps from the backend ( like email notification)  (This happens when we click on assign button without checking the books from UI resulting the booklist will be empty)
    
2. if booklist is not empty and action is "assign"
URL - /v1/autographamt/projects/assignments
Header Request - {"projectId":"2","userId":3,"books":["3jn","2jn","1jn"],"action":"assign"}
Response - {"success":true, "message":"Books assigned"}

If the booklist is not empty:
-check in the database if there is any book assigned to this user,
a. If YES, 
     Keep(save in the db) the books that are in the booklist and remove all other books from the database.

b. If NO,
    Save/Keep the book in database that's in the booklist

kavitharaju:
Is it required that we need different messages saying books assigned and books unassigned?
Can both of those be conveyed with one message saying "Book assignments updated"?

Because I see, these two operations are not mutually exclusive. Both adding new assignments and removing some of old assignments can happen in same request as we use same action "assign" now.

kavitharaju:
Let me try to sum up the complete requirements for this API

API endpoint: "/v1/autographamt/projects/assignments"
method: POST
Inputs expected: `userId`, `projectId`, `books`(list of bookcodes, can be empty), `action`(either add_user or assign)
Success responses:
1. `{"success":true, "message":"User added"}`
2. `{"success":true, "message":"Book assignments updated"}`

Failure responses:
1. `{"success":false, "message":"User already added to project"}` for `add_user` action
2. `{"success":false, "message":"User not added to project yet"}` for `assign` action
3. `{"success":false, "message":"No change in book assignments"}` for `assign` action

Functions Performed
1. **add_user**: check and ensure that the user-project pair do not exists in the assignments table and add a new entry to it.   Book list can be empty or can have list of book codes
2. **assign**: check and ensure that the user-project pair exists in the assignments table. Also checks  and ensures that the new booklist is not same as what we have in DB and **replace** already present books with the new list. Book list can be empty or can have list of book codes

@Joel-C-Johnson  Would this cover your needs of proper response messages and all assign and unassign actions you need from the API?

Joel-C-Johnson:
Yeah I hope this will work fine as we excepted

kavitharaju:
I have updated as per above API description

kavitharaju:
@Joel-C-Johnson 
Is there a reason this is not approved or merged yet? If this is not required anymore we can close it for cleaning up the issues as we have more to new version's developement now.

Joel-C-Johnson (approved):
This works well.  Thank you @kavitharaju .  Sorry for the delay from my end

jincypjose (approved):

PR #122, opened 2021-03-03
Modify upload bible call as per new json structure

RevantCI:
Hi Kavitha, 
Have sent a PR as per our discussion for the upload bible new json structure.
Let me know if there are any changes required. 
Will test the bible get calls and send any changes requried there as well.

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -1412,11 +1412,10 @@ def parseDataForDBInsert(usfmData):
 		for verse in verseData:
 			crossRefs = ""
 			footNotes = ""
-			if 'metadata' in verse and 'cross-ref' in verse['metadata']:
-				crossRefs = verse['metadata']['cross-ref']
-			if 'metadata' in verse and 'footnote' in verse['metadata']:
-				footNotes = verse['metadata']['footnote']
-			verseNumber = verse.get('verseNumber',"").strip()
+			verseNumber = verse.get('verseNumber',"").strip() if isinstance(verse, dict) else ""
+			if not verseNumber:
+				# not a verse
+				pass
 			if normalVersePattern.match(verseNumber):
 				verseText = verse["verseText"]
 				dbVerseText = verseText
@@ -1465,7 +1464,7 @@ def parseDataForDBInsert(usfmData):
 					dbInsertData.append((ref_id, "", "", ""))
 					verseContent.append('')
 
-			elif verseNumber:
+			else:
 				log.info("Unrecognized pattern in %s chapter %s verse %s",bookName, chapterNumber, verseNumber)
 	return dbInsertData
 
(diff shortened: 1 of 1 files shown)


agmt/main.py
```diff
@@ -1401,24 +1401,24 @@ def parseDataForDBInsert(usfmData):
...
+		verseData = chapter["contents"]
 		for verse in verseData:
 			crossRefs = ""
 			footNotes = ""
 			if 'metadata' in verse and 'cross-ref' in verse['metadata']:
 				crossRefs = verse['metadata']['cross-ref']
 			if 'metadata' in verse and 'footnote' in verse['metadata']:
 				footNotes = verse['metadata']['footnote']
-			verseNumber = verse['number'].strip()
+			verseNumber = verse.get('verseNumber',"").strip()
```
kavitharaju, line 1419: I think you may need to do a check before accessing the `verseNumber`. Because `chapter.contents` can be verses or other objects like `\p`. So first check if `verseNumber` is in the object to make sure it is a verse object.

agmt/main.py
```diff
@@ -1401,24 +1401,24 @@ def parseDataForDBInsert(usfmData):
...
-		chapterNumber = chapter["header"]["title"]
-		verseData = chapter["verses"]
+		chapterNumber = chapter["chapterNumber"]
+		verseData = chapter["contents"]
 		for verse in verseData:
 			crossRefs = ""
 			footNotes = ""
 			if 'metadata' in verse and 'cross-ref' in verse['metadata']:
 				crossRefs = verse['metadata']['cross-ref']
 			if 'metadata' in verse and 'footnote' in verse['metadata']:
```
kavitharaju, line 1417: I dont think the new structure will have `metadata`. So these `if` blocks will be unreachable code while using  new JSON structure. Its not an issue, to leave it like that. Just noting it down, thats all

kavitharaju (commented):
In our offline discussion we had talked about `keeping the code so as to handle both kinds of JSON structures. I see you are not doing that here. With these changes only the new strcuture will be accepted. I hope that is done intentionally

agmt/main.py
RevantCI: Yes, that is why I am using get and have set the defaut value as ""  in case it doesnt find a `verseNumber` ie it's a `\p`.
Have added a check in last elif to check for empty string as well so it passes through without doing anything for `\p` objects etc.So it will not match any if condition and not do anything in that for loop.
if you want can put `if not verseNumber` condition in the starting of the if elif blocks so that the code is more readlble that we are considering that case for anyone maintaing the code.

kavitharaju:
> In our offline discussion we had talked about `keeping the code so as to handle both kinds of JSON structures. I see you are not doing that here. With these changes only the new strcuture will be accepted. I hope that is done intentionally.
> 
> I understand that we are doing this for the new requirement for ESV bibles. But for testing, I would recommend to test not only with ESV but also the new JSONs obtained by the IRV usfms we have uploaded before, if that is not too much of work

@RevantCI 
Are these okay?

kavitharaju (approved):
I see you have uploaded all 12 GLs and English bibles in sandbox. So I assume that is enough to confirm this PR is fine. The changes required in other Apps can be dealt with separately. So I am going to approve this PR

samuelanthrayose (approved):

PR #129, opened 2021-04-06
Translation APIs: Base Implementation

kavitharaju:
First PR for AGMT and other Translation APIs


joelthe1 (approved):
LGTM 👍 
Putting this here for future work that we should clean-up `requirements.txt`

PR #143, opened 2021-05-11
Normalize translations field

### LINKED ISSUE #138: Normalize translation field in DB
The `translations` field in `Translation_memory` table is currently a multi-valued field.
split it into two fields `translation` and `frequency`, allowing multiple rows for each sense/translation of a source language token.

This would allow us to efficiently query based on `translations` field as well, enabling us to search for a token and its known translations in the reverse manner(target to source) also.

### LINKED ISSUE #140: Fuzzy token matching
Use fuzzy string matching for fetching similar tokens and translations from DB for suggestions.
For implementing that add a new field, with romanized source token. And index this with soundex indexing and use soundex and levenshtein to identify similar tokens.

Do we need another field with romanized translation, allowing reverse search?.

Refer [this article](https://blog.crunchydata.com/blog/fuzzy-name-matching-in-postgresql) for implementation.

### LINKED ISSUE #142: Shorten requirements.txt file
Include only the packages that are directly imported in code and not their dependencies.

kavitharaju:
* The `translation_details` field in `translation_memory` table which had the type `{"value1": {frequency: 1}, "value2": {"frequency":0}}` has been split into 2 fields `translation`, `frequency` . All the code updations for this transitions. closes #138 
* Supports fuzzy string matching for tokens(and translations) while finding translation suggestions from translation memory. Adds new fields, `source_token_romanized` and `translation_romanized` which are indexed with soundex. Also uses soundex based filtering and levinshetein based scoring to identify close matches while querying. closes #140 
* The requirements.txt file which was earlier created using `pip freeze` command is removed. Added a manually created file which includes only those packages that are directly imported in code and not their dependencies. closes #142


app/crud/nlp_crud.py
```diff
@@ -685,44 +623,66 @@ def add_to_translation_memory(db_, src_lang, trg_lang, gloss_list, default_val=0
...
+                updated_args['metaData'] = gloss['tokenMetaData']
+                token_row = db_models.TranslationMemory(**updated_args)
+            db_.add(token_row)
+            db_content.append(token_row)
     db_.commit()
+    result_dict = {}
     for item in db_content:
         db_.refresh(item)
-    return db_content
+        if item.token not in result_dict:
```
joelthe1, line 678: Just making sure that the logic here should indeed be `if` and not `elif`.

app/crud/nlp_crud.py
```diff
@@ -685,44 +623,66 @@ def add_to_translation_memory(db_, src_lang, trg_lang, gloss_list, default_val=0
...
+            result_dict[item.token] = {"token":item.token, "translations":{}}
+        if item.translation is not None:
```
joelthe1, line 680: Just making sure that the logic here should indeed be `if` and not `elif`.

app/crud/utils.py
```diff
@@ -135,3 +136,10 @@ def parse_usfm(usfm_string):
         raise TypeException(stderr.decode('utf-8'))
     usfm_json = json.loads(stdout.decode('utf-8'))
     return usfm_json
+
+
+def to_eng(data):
```
joelthe1, line 141: What happens if this function throws an error?

kavitharaju: If you are asking about `to_eng()` function, as it is not handled specifically, would be the handled by `any_exception_handler()` defined in main. It would return the error message to user with "Internal server error" title and status code `500`

app/crud/nlp_crud.py
kavitharaju: No, it is not `elif`. Now, as per the updated translation memory table structure, we can have multiple rows for same token with different translations. But in the response object we are need just one entry for one token with all available translations as an array in it. Here first `if` is for adding the token to return object upon seeing it for the first time and second `if` is for adding translations into the token object, which should happen on first occurrence as well as subsequent occurrences.

Issue #150, opened 2021-05-20
Need to upload only usfm for bible data upload


kavitharaju:
For the api `POST /v2/bibles/{source_name}/books` now user needs to upload the JSON obtained from usfm-grammar, along with the USFM, to the server. Use usfm-grammar at server side and ask user only for USFM.

kavitharaju:
@joelthe1 says, we should allow the user the option to also upload a USFM-Grammar formatted JSON as input instead of USFM as an input format.

Issue #147, opened 2021-05-19
Allow partial word search of language api using query parameter


RevantCI:
- Add a query parameter - like **search** or **query** to the language get call for v2 vachan-api 
- This is to facilitate searching of langauges when loading the list of languages in the dropdown as the number is quite high
- To use server side pagination it needs to fiter them also  at the server side using the query parameter
- It can search language_name and language_code partially using wildcard characters. at the ends

PR #151, opened 2021-05-21
Feature/search language

kavitharaju:
This PR includes the following major changes
- newly added language codes are validated 
- languages are indexed and there is a search query parameter added which does fuzzy matching on multiple fields of languages #147 
- only USFM needs to be provided by user to add books to bible, and not the JSON. #150


joelthe1:
I also think we should allow users to upload JSON (the output of USFM-Grammar) to add books.

kavitharaju:
> I also think we should allow users to upload JSON (the output of USFM-Grammar) to add books.

@joelthe1 You mean, we should allow them to upload JSON also, optionally, along with USFM?

joelthe1:
> @joelthe1 You mean, we should allow them to upload JSON also, optionally, along with USFM?
I meant we should allow the user the option to also upload a USFM-Grammar formatted JSON as input instead of USFM as an input format.

Issue #173, opened 2021-07-02
Cannot highlight untranslated tokens on occurcepanel


kavitharaju:
When starting a translation, all the tokens in a sentence are untranslated. Then the `draftMeta` would be like 
`[[0,length], [0, length], "untranslated"]`
Then for a token with occurance  offset say [5,10], there is no excat segment in draftMeta. But as sentence=draft, the offset would be same in draft also.

But after we apply a translation for the first token from `[0,4]`, the draft length may vary and then it will become difficult to determine the exact offset in draft to highlight it.

PR #174, opened 2021-07-02
Agmt highlightable sentence drafts

kavitharaju:
To resolve the issue #173
New api `/v2/autographa/project/token-sentences` that takes a token and its required occurrence list.
The source sentences and draft would be returned as in get project source. But it would be ensured that the draftMeta will have segements that covers the specified occurences to be able to easily determine and hight the translations of the token in draft.


kavitharaju:
@Joel-C-Johnson using this API it can be garrenteed that the offset that matches the token would be present in normal cases. But I would recommend that you do a set intersection(check if offset is a subset of segment in draftMeta) to find the matching segment instead of doing exact match.

This is to cover cases like this, suppose there is  a confirmed tranlastion for the token `Son of God` in the verse we are looking at.
But after another tokenization suppose we check for the translation of token `God` then `draftMeta` will only have segment corresponding to `son of god` and not `god`.

PR #179, opened 2021-07-07
Feature/mutations

sijumoncy:
This branch have the Mutations created for ADD and UPDATE features of Languages.


app/crud/utils.py
```diff
@@ -133,6 +133,8 @@ def parse_usfm(usfm_string):
                          stderr=subprocess.PIPE,
                          shell=True)
     stdout, stderr = process.communicate()
+    print("--STDOUT----=",stdout)
```
kavitharaju, line 136: These prints added while debugging or development could be removed

app/graphql_api/mutations.py
```diff
@@ -1,10 +1,80 @@
...
+class UpdateLanguage(graphene.Mutation):
+    class Arguments:
+        language_updateargs = InputUpdateLang(required=True)
+
+    language = graphene.Field(types.Language)
+
+    def mutate(root,info,language_updateargs):
+        schema_model = utils.convert_graphene_obj_to_pydantic(language_updateargs,schemas.LanguageEdit)
+        result = structurals_crud.update_language(db_=next(get_db()),lang=schema_model)  
+        print("==============================================")
```
kavitharaju, line 64: We could remove print statments like this to keep code base cleaner

kavitharaju (changes requested):
Adding few points that struck me at first glance. Shall run and test locally and respond. You can take a look at these in the mean while.

Are new files `app/tkn_offset[0]` and `edited parts.txt` required?

Did you use pylint? It shows warnings(see the bottom of PR page), because of trailing spaces and multiple new lines at the end of file.

app/graphql_api/mutations.py
```diff
@@ -1,10 +1,80 @@
...
+    def mutate(root,info,language_Addargs):
+        schema_model = utils.convert_graphene_obj_to_pydantic(language_Addargs,schemas.LanguageCreate)
+        result =structurals_crud.create_language(db_=next(get_db()),lang=schema_model)
+        language = types.Language(
+            language = result.language,
+            code = result.code,
+            scriptDirection = result.scriptDirection,
+            metaData = result.metaData
+        )
+        return AddLanguage(language = language)
```
kavitharaju, line 43: Can we also return a message stating if the operation was successful or like in REST's response

app/graphql_api/mutations.py
```diff
@@ -1,10 +1,80 @@
...
+import schemas
+
+
+############ ADD NEW Language #################
+""" ADD Language Input """
+class InputAddLang(graphene.InputObjectType):
+    language = graphene.String()
+    code = graphene.String(description="language code as per bcp47(usually 2 letter code)")
+    scriptDirection = graphene.String()
+    metaData = graphene.JSONString()
```
kavitharaju, line 21: May be a description would be good here, stating we expect a dict like value as it is not very self explanatory

app/graphql_api/mutations.py
```diff
@@ -1,10 +1,80 @@
...
+
+
+class AddLanguage(graphene.Mutation):
+    class Arguments:
+        language_Addargs = InputAddLang(required=True)
```
kavitharaju: I see we are expecting an input object here instead of individual parameters like you mentioned in our discussion. Is there a reason for doing it this way?

app/graphql_api/mutations.py
```diff
@@ -28,39 +33,44 @@ class Arguments:
...
-    language = graphene.String()
-    code = graphene.String(description="language code as per bcp47(usually 2 letter code)")
-    scriptDirection = graphene.String()
-    metaData = graphene.JSONString()
+    languageId = graphene.Int(required=True)
+    language = graphene.String(required=True)
+    code = graphene.String(required=True,\
+        description="language code as per bcp47(usually 2 letter code)")
+    scriptDirection = graphene.String(required=True)
+    metaData = graphene.JSONString(description="Expecting a dictionary Type")
```
kavitharaju, line 65: I think we need to keep only language Id required here. That is how it is in REST right?
image

app/graphql_api/types.py
```diff
@@ -19,7 +19,7 @@ class ContentType(graphene.ObjectType):
 
 class Language(graphene.ObjectType):
     '''output object for Language'''
-    #language_id = graphene.Int()
+    language_id = graphene.Int()
```
kavitharaju, line 22: We use camel casing as much as possible in the output, to stick to the JS practices. As our clients are mostl in JS

app/test/test_graphql_languages.py
```diff
@@ -0,0 +1,47 @@
...
+        code
+        scriptDirection
+        metaData
+    }
+    }
+    """
+    executed = client.execute(default_get_query)
+    assert isinstance(executed, Dict)
+    assert len(executed["data"]["languages"])>0
+    assert isinstance(executed["data"]["languages"], list)
```
kavitharaju, line 34: we also need to test the queried fields, like language, code etc, are in the result.

app/test/test_graphql_languages.py
```diff
@@ -0,0 +1,47 @@
...
+    """test for get only 1 field data with limit parameter"""
+    default_get_query = """
+        {
+    languages(limit:1){
+        language
+    }
+    }
+    """
+    executed = client.execute(default_get_query)
+    assert executed == {"data": {"languages": [{"language": "Afar"}]}}
```
kavitharaju, line 47: This is a risky test. We cannot always guarantee this particluar language would the first returned. It will depend on the order languages were inserted in DB. Better query for this  language specifying language name or code

kavitharaju (changes requested):

app/graphql_api/mutations.py
```diff
@@ -1,10 +1,97 @@
...
 from dependencies import get_db
+#pylint: disable=E0611
 from graphql_api import types, utils
+import schemas
+
+#pylint: disable=R0901,too-few-public-methods
+class Output(graphene.ObjectType):
+    """Language output type"""
+    msg = graphene.String()
+    language_type = graphene.Field(types.Language)
```
kavitharaju, line 18: Its would be better to name it differently that just `Output`. There is going to be more output objects for each of the API group. So follow a naming pattern that can be adopted throughout

app/graphql_api/mutations.py
```diff
@@ -1,10 +1,97 @@
...
+
+############ ADD NEW Language #################
+class InputAddLang(graphene.InputObjectType):
+    """ADD Language Input"""
+    language = graphene.String(required=True)
+    code = graphene.String(required=True,\
+        description="language code as per bcp47(usually 2 letter code)")
+    scriptDirection = graphene.String(required=True)
+    metaData = graphene.JSONString(description="Expecting a dictionary Type")
```
kavitharaju, line 27: Is scriptDirection a required field in REST?

app/'
```diff
@@ -0,0 +1 @@
+Unable to initialize device PRN
```
kavitharaju, line 1: If this new file used for anything?

app/graphql_api/types.py
kavitharaju: I see you changed it to languageId. Is it working as expected now?

app/'
sijumoncy: I think its automatically created and none of these files have any connection with the feature/mutations

kavitharaju: Oh, is it? Okay then. I am going to approve and merge this PR. We can check why these files are being created and also add more tests for these mutations as the next task

PR #180, opened 2021-07-07
fix offset calculation issues in tokenization

kavitharaju:
There was on test case failing in suggestions module. This fix resolves that


app/crud/nlp_crud.py
```diff
@@ -113,7 +113,7 @@ def tokenize(db_:Session, src_lang, sent_list, use_translation_memory=True, incl
                     lngst = memory_trie.longest_prefix(key)
                     if lngst.key is not None:
                         new_chunks.append("###"+lngst.key.replace('/',' '))
-                        temp = temp[len(lngst.key):]
+                        temp = temp[len(lngst.key)+1:]
```
joelthe1, line 116: For the future- I have found there is usually a better name that can be used instead of `temp`.

PR #185, opened 2021-07-12
Bugfix/graphql tests commiting to db

### LINKED ISSUE #184: GraphQL mutations commiting changes
Normal test cases for REST apis are written so that the TestClient overrides the normal DB connection and starts a new transaction which would be rolled back after each test function. Thus the PUT or POST apis do not make lasting changes on the DB. So running the same testcase creating the same item again next time will not throw an error.

The tests for graphQL Mutations should also work as above. But the current way the graphQL app is configured in fast API we are not able to use this alternate DB connection mechanism effectively.

kavitharaju:
Fixes #184 

The DB connection for REST apis are passed as Fastapi dependancy and is overriden by an alternate connection in testClient.

The graphene App or test Client doesn't have the concept of dependancies and hence it is hard to pass on the same.
This PR re-structures how grapheneApp is connected to the main fastAPI app. It is now following a method that is very close to the routers for REST and define a POST and GET endpoints for all the graphQL requests. These routers are taking the DB connection as a dependancy and passing it on to the graphene module as a request context which the graphene understand.
Doing this allows us to override the DB connection dependancy in Testclient.

This also includes changes to the graphQL tests as per this change in configuration. It doesn't use the graphene's test client or `execute` method for making gql requests. Instead gql queries are passed on as regular http POST requests to the endpoint `/graphql`

### DIFF
--- a/app/'
+++ b/app/'
@@ -1,1 +1,0 @@
-Unable to initialize device PRN
--- a/app/graphql_api/__init__.py
+++ b/app/graphql_api/__init__.py
@@ -3,6 +3,9 @@ from fastapi import APIRouter, Depends
 import graphene
 from starlette.requests import Request
 from starlette.graphql import GraphQLApp
+
+#pylint: disable=E0401
+#pylint gives import error if relative import is not used. But app(uvicorn) doesn't accept it
 from graphql_api import queries, mutations
 from dependencies import get_db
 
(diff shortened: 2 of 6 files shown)


joelthe1:
This looks okay to me, though I would just point out what the [spec](https://graphql.org/learn/serving-over-http/#post-request) recommends when using POST requests about handling a couple extra cases which I am unable to verify, if those are done.

kavitharaju:
Yes.
running GET `127.0.0.1:8000/graphql?query={languages{language}}` and 
POST with body=graphQL from Postman was successfully handled. So I think both those cases are being handled.

joelthe1:
Okay, I assume `graphene` already conforms to these recommendations.

PR #203, opened 2021-07-16
Bugfix/translation suggestions

### LINKED ISSUE #193: Cannot customize Stopwords and punctuations in AgMT
AgMT project shows provision for using user-specified stopwords list and punctuations list so that tokenizations can be controlled.
But this feature is not working as expected

### LINKED ISSUE #194: Server error in suggestions module
API : autographa/project/suggestions?project_id=xxx
Gives error: 
`{'error': 'Error', 'details': "'dict' object has no attribute 'stopwords'"}`

The API worked as expected before. After a new custom stopword list was added to the project, this API gave this error

kavitharaju:
- fixes #193 
- fixes #194 
- Adds an end-to-end test for translation workflow
- Uses word(non-phrase) tokenization within the suggest-translation operation

### DIFF
--- a/app/crud/nlp_crud.py
+++ b/app/crud/nlp_crud.py
@@ -498,6 +498,7 @@ def alignments_to_trainingdata(db_:Session, src_lang, trg_lang, alignment_list,
     sugg_json = {item[0]:item[1] for item in new_trie.items()}
     json.dump(sugg_json, output_file, ensure_ascii=False)
     output_file.close()
+    log.warning("trie model saved on server: %s", output_path)
     new_trie = rebuild_trie(db_, src_lang, trg_lang)
     suggestion_trie_in_mem[src_lang+"-"+trg_lang] = new_trie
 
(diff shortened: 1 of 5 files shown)


app/crud/nlp_crud.py
```diff
@@ -498,6 +498,7 @@ def alignments_to_trainingdata(db_:Session, src_lang, trg_lang, alignment_list,
     sugg_json = {item[0]:item[1] for item in new_trie.items()}
     json.dump(sugg_json, output_file, ensure_ascii=False)
     output_file.close()
+    log.warning("trie model saved on server: %s", output_path)
```
joelthe1, line 501: I would think this is more appropriate to have `log.info("")`. What do you think?

kavitharaju: while on server the logging level would be set to Warning. So only warning or error logs would be written.

The reason I am logging this is to identify and remove any wrong or incorrect alignments added to server. Hence used warning

Issue #196, opened 2021-07-16
Mutations for Commentary APIs


kavitharaju:
- implement mutations corresponding to PUT and POST methods
- implement tests for mutations and queries

Issue #195, opened 2021-07-16
Mutations for Bibles APIs


kavitharaju:
- implement mutations corresponding to PUT and POST methods
- implement tests for mutations and queries

PR #205, opened 2021-07-21
Feature/mutations

sijumoncy:
#195 #196 

PR for mutation and test of GraphQL bible-api and commentaries

### DIFF
--- a/app/graphql_api/mutations.py
+++ b/app/graphql_api/mutations.py
@@ -3,7 +3,7 @@ import graphene
 
 #pylint: disable=E0401
 #pylint gives import error if relative import is not used. But app(uvicorn) doesn't accept it
-from crud import structurals_crud
+from crud import structurals_crud,contents_crud
 #pylint: disable=E0611
 from graphql_api import types, utils
 import schemas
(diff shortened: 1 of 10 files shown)


app/graphql_api/mutations.py
```diff
@@ -358,6 +358,295 @@ def mutate(self,info,source_arg):
         message = "Source edited successfully"
         return AddSource(message=message,data=source_var)
 
+########## Add Bible books ########
+class InputAddBible(graphene.InputObjectType):
+    """Add Bible Input"""
+    source_name = graphene.String(required=True,\
+        description="pattern: ^[a-zA-Z]+(-[a-zA-Z0-9]+)*_[A-Z]+_\\w+_[a-z]+$")
+    books = graphene.List(graphene.JSONString,required=True,\
```
kavitharaju, line 366: Need to specify the expected inner contents as USFM and/or JSON. I suggest you use a separate InputObject like the `BibleBookUpload` used in REST

app/graphql_api/mutations.py
```diff
@@ -358,6 +358,295 @@ def mutate(self,info,source_arg):
...
+        message = "Bible books uploaded and processed successfully"
+        return AddBible(message=message,data=bible_content_list)
+
+########## Edit Bible books ########
+class BibleEditDict(graphene.InputObjectType):
+    """bible books inputs"""
+    book_code = graphene.String(
+        description="pattern:^[a-zA-Z1-9][a-zA-Z][a-zA-Z]$")
+    USFM = graphene.String(description="USFM Data")
+    JSON = graphene.JSONString(description="JSON Data")
```
kavitharaju, line 409: Not any JSON would be accespted here. Only JSON structure obtained from USFM-Grammar or one like that would be accepted. So its good to specify that too in description

app/test/test_gql_commentaries.py
```diff
@@ -0,0 +1,842 @@
...
+        "verseEnd":1, "commentary":"first verse of Genesis"}
+        ]
+    }
+    }
+    executed8 = gql_request(query=ADD_COMMENTARY,operation="mutation",variables=variable8)
+    assert "errors" in executed8.keys()
+
+def test_get_after_data_upload():
+    '''Add some data into the table and do all get tests'''
+    variable = {
```
kavitharaju, line 245: Are we running a common set of sets for all GET requests(queries) to check for the correct functioning of skip, limit etc as in tests for REST?

kavitharaju (changes requested):

app/graphql_api/mutations.py
```diff
@@ -358,6 +358,301 @@ def mutate(self,info,source_arg):
         message = "Source edited successfully"
         return AddSource(message=message,data=source_var)
 
+########## Add Bible books ########
+class InputBibleDict(graphene.InputObjectType):
+    """Add Bible Dict"""
+    USFM = graphene.String()
+    JSON = graphene.JSONString()
```
kavitharaju, line 365: The same `description="Provide JSON structure obtained  from USFM-Grammar or one like that"` can be given here as well

app/test/__init__.py
```diff
@@ -109,20 +109,20 @@ def check_soft_delete(unit_url, check_post, data, delete_data):
...
     assert isinstance(executed, Dict)
-    if len(executed["data"]["languages"]) > 1:
+    if len(executed["data"][api_name]) > 1:
         executed2 = gql_request(query2)
         assert isinstance(executed2, Dict)
-        assert executed["data"]["languages"][1] == executed2["data"]["languages"][0]
+        assert executed["data"][api_name][1] == executed2["data"][api_name][0]
 
-def check_limit_gql(query,limit):
+def check_limit_gql(query,limit,api_name):
```
kavitharaju, line 121: What I would suggest here is define multiple checks with different values for skip and limit in these functions itself. See how multiple cases are tested in corresponding check functions of REST. Then you will not have to multiple write long queries and call this function multiple time for each test file. Construct it such that it is called only once and pass in a minimal query which is then appended with skip and limit parameters here. That would save us a lot of code.

Also these test are to be done for all queries, not just languages and commentaries. Add them in all tests you have written before

app/test/test_gql_bibles.py
```diff
@@ -0,0 +1,1081 @@
...
+  for item in items:
+      assert_positive_get_for_books(item)
+      assert "USFM"  in item
+      assert "JSON" in item
+      assert "audio" in item
+      assert item["audio"] is None
+
+  #filter with book code 
+  query2 = """
+      {
```
kavitharaju, line 571: I see you are asking for all fields in all get queries. Its good to add them in one query, later on you can just query for minimal fileds to make sure the function is working as expected. I think that would save lot of line-of-code in these test files and make it more readable

kavitharaju (changes requested):

app/test/__init__.py
```diff
@@ -109,20 +109,49 @@ def check_soft_delete(unit_url, check_post, data, delete_data):
...
-    if len(executed["data"]["languages"]) > 1:
+    if len(executed["data"][api_name]) > 1:
         executed2 = gql_request(query2)
         assert isinstance(executed2, Dict)
-        assert executed["data"]["languages"][1] == executed2["data"]["languages"][0]
+        assert executed["data"][api_name][1] == executed2["data"][api_name][0]
 
-def check_limit_gql(query,limit):
+    # fetch a non existant page, with skip and limit values
+    query3 = query.replace("arg_text","skip:50000,limit:10")
```
kavitharaju, line 124: Having a place holder and replacing would work too. 
But can't we pass it as variables?

sijumoncy: means?
query = """
  getsomething(sourcename:"something",variable){
        data
}
"""

like this ..?

kavitharaju: I meant using the variables as allowed by GQL format instead of hard-coding the skip and limit values.
```
query = """
        query something($skip:int, $limit:int){
        something(skip:$skip){
           ...
        }
        }
"""
```
and passing 
```
variables = {
     "$skip": 0,
     "$limit":0
}
```
to ` gql_request(query=query, operation="query", variables=variables)`

sijumoncy: i will look into it

kavitharaju (approved):

Issue #202, opened 2021-07-16
Tests for Translation Queries


kavitharaju:
Make sure we have GraphQL endpoints corresponding to all operation under the sections
Autographa-Translation, Translation Suggestion and Generic Translation. Note that many of the PUT methods are implemented as Queries in GraphQL.

Add tests for already added queries and any new mutations in these section of APIs

PR #213, opened 2021-07-29
Autographa translation gql

sijumoncy:
#202 

PR contain mutation and test related to the graphql translation suggestions and generic translation


app/graphql_api/mutations.py
```diff
@@ -1145,3 +1239,5 @@ class VachanMutations(graphene.ObjectType):
     edit_agmt_project = EditAGMTProject.Field()
     create_agmt_project_user = AGMTUserCreate.Field()
     edit_agmt_project_user = AGMTUserEdit.Field()
+    apply_token_translation = TokenApply.Field()
```
kavitharaju, line 1242: These mutations are belonging withing the "autographa" section right? Better name the with "agmt" as you have done in the previous cases to differentiate them from generic translation operatios

kavitharaju: Please make sure the queries implemented before in Autographa translation and generic translation are also named appropriately that will allow us to differentiate them. All autographa project specific operations should have "agmt" in the name.

kavitharaju (changes requested):

app/graphql_api/mutations.py
```diff
@@ -1117,6 +1117,100 @@ def mutate(self,info,info_arg):
...
+            draftMeta = item.draftMeta
+            )
+            dict_content_list.append(comm)
+        message = "Token translations saved"
+        return TokenApply(message=message,data=dict_content_list)
+
+# Get Token Sentances
+class InputGetSentance(graphene.InputObjectType):
+    """Inputs for Aplly Token"""
+    project_id = graphene.Int(required=True)
```
kavitharaju, line 1177: Should this operation be a mutation or query? This was an API that was added later. So might not have been covered when we did queries. But I think it should be a query as it doesn't make any change on the server

app/graphql_api/mutations.py
```diff
@@ -1117,6 +1117,100 @@ def mutate(self,info,info_arg):
...
+    token = graphene.String(required=True)
+    occurance = graphene.List(InputOccurance,required=True)
+
+class GetTokenSentance(graphene.Mutation):
+    "Mutations for  Get Token Sentance "
```
kavitharaju, line 1182: Should this operation be a mutation or query? This was an API that was added later. So might not have been covered when we did queries. But I think it should be a query as it doesn't make any change on the server

app/graphql_api/mutations.py
sijumoncy: sure

app/graphql_api/mutations.py
sijumoncy: In the Rest API its shown as PUT , so I done it as a Mutation

app/graphql_api/mutations.py
```diff
@@ -1117,6 +1117,60 @@ def mutate(self,info,info_arg):
...
+    """Input object for applying token translation"""
+    sentenceId = graphene.Int(required=True)
+    offset = graphene.List(graphene.Int,required=True,\
+        description="Max-Min Item 2")
+
+class InputToken(graphene.InputObjectType):
+    """Input object for applying token translation"""
+    token = graphene.String(required=True)
+    occurrences = graphene.List(InputOccurance,required=True)
+    translation = graphene.String(required=True)
```
kavitharaju, line 1132: I see new classes `InputOccurance` and `InputToken` defined here. Just wondering are there older classes defined in mutations or quries that could be used here.

app/graphql_api/queries.py
```diff
@@ -220,6 +220,17 @@ def resolve_agmt_project_token_translation(self, info, project_id, sentence_id,
         return projects_crud.obtain_agmt_token_translation(db_, project_id, token=None,
             occurrences=occurrences)[0]
 
+    #Agmt get token Sentance
+    agmt_get_token_sentence = graphene.List(types.Sentence,
+        description="Query the translation done for sentance in an AgMT project",
+        project_id=graphene.ID(required=True),token=graphene.String(required=True),
+        occurrences = graphene.List(types.TokenOccurenceInput,required=True))
```
kavitharaju, line 227: This corresponds to the REST API https://api.vachanengine.org/docs#/Autographa-Translation/get_token_translation_v2_autographa_project_token_translations_get right? Why is it names `agmt_get_token_sentence` and not `agmt_get_token_translation`? 
Also, the naming it `agmt_project_token-translation` would be more in line with the other query names as we havn't used get in any other query names

sijumoncy: just look this 
token_tranlate

kavitharaju: yes, that is a different API. The current change looks good

app/graphql_api/mutations.py
kavitharaju: In REST, we are not allowed to have body params in GET. Here we needed to pass occurence_list as body param, so implemented it as PUT just for that. It should be query in GraphQL. The change you did looks good.

kavitharaju (approved):

Issue #201, opened 2021-07-16
Translation Suggestion


kavitharaju:
- implement mutations corresponding to learn/gloss and learn/alignment methods
- implement tests for mutations and queries

PR #214, opened 2021-08-02
Translation suggetion gql

sijumoncy:
#201 

PR Contains tests related with translation suggestions

### DIFF
--- a/app/graphql_api/mutations.py
+++ b/app/graphql_api/mutations.py
@@ -1285,6 +1285,7 @@ class AddGloss(graphene.Mutation):
         dict_content_list = []
         for item in result:
             #pylint: disable=R1726
+            #pylint: disable=E0012
             if "translations" and  "metaData" in item:
                 dict_var = types.Gloss(
                token = item["token"],
@@ -1341,6 +1342,7 @@ class AddAlignment(graphene.Mutation):
         dict_content_list = []
         for item in result:
             #pylint: disable=R1726
+            #pylint: disable=E0012
             if "translations" and  "metaData" in item:
                 dict_var = types.Gloss(
                token = item["token"],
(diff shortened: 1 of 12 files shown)


app/graphql_api/types.py
```diff
@@ -284,7 +284,7 @@ class GlossInput(graphene.InputObjectType):
     token = graphene.String(description='example="love"', required=True)
     translations = graphene.List(graphene.String,
         description="example=['प्यार', 'प्रेम', 'प्रेम करना']")
-    tokenMetaData = Metadata(description='example={"word-class":["noun", "verb"]}')
+    tokenMetaData = graphene.JSONString(description='example={"word-class":["noun", "verb"]}')
```
kavitharaju, line 287: I see you are not using the Metadata class but using the JSONString type for metadata field. Which is good.
But are we following the same pattern in other places too, like in the metadata filed of projects, in dictionary tables, in versions table etc? I would suggest we follow same pattern through out

app/graphql_api/mutations.py
```diff
@@ -1190,3 +1389,7 @@ class VachanMutations(graphene.ObjectType):
     create_agmt_project_user = AGMTUserCreate.Field()
     edit_agmt_project_user = AGMTUserEdit.Field()
     agmt_apply_token_translation = AgmtTokenApply.Field()
```
kavitharaju, line 1391: I know I suggested this name. But seeing the pattern you name other mutations, I think it would be good to have this also start with the verb and have agmt in the middle, as in `applyAgmtTokenTranslation`. I think that would go along with the other names you gave

app/graphql_api/types.py
sijumoncy: Throughout mutation used metadata as JSONString

kavitharaju (approved):

PR #217, opened 2021-08-05
Fix wrong SQL in tokenization

### LINKED ISSUE #216: Database error on tokenizing hindi text
The tokenization APIs `/v2/translation/tokens` and `/v2/autographa/project/tokens` giving error
```
{
  "error": "Database Error",
  "details": "syntax error at or near \"ii\"\nLINE 7: ...soundex(source_token_romanized) = soundex('piiddh'ii huii') ...\n                                                             ^\n"
}
```
The error is observed when the input sentence given for tokenization has the word "पीढ़ी".

kavitharaju:
fixes #216 

The syntax error given by the database was due to the apostrophe in the romanized word injected in the SQL query. 
As we require to run query using soundex, the ORM query builder's `text()` feature is used to inject this part of the query. The romanization library we use, unidecode, adds apostrophes(`'`) within words which was resulting in the error. 

In this fix we remove apostrophes from unidecode's response.

### DIFF
--- a/app/crud/nlp_crud.py
+++ b/app/crud/nlp_crud.py
@@ -726,7 +726,7 @@ def get_gloss(db_:Session, index, context, source_lang, target_lang): # pylint: 
         db_models.TranslationMemory.token,
         db_models.TranslationMemory.translation,
         db_models.TranslationMemory.frequency,
-        text("levenshtein(source_token,'%s') as lev_score"%word)
+        text("levenshtein(source_token,:word) as lev_score").bindparams(word=word)
         ).filter(
             db_models.TranslationMemory.source_language.has(code=source_lang),
             db_models.TranslationMemory.target_language.has(code=target_lang),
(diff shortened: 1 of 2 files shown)


joelthe1:
I think we should sanitize all input to SQL queries so that we don't have any possibility of SQL injection attacks.

kavitharaju:
I have changed the way we add the romanized words to the query text. 

For most of the query parts we used SQLAlchemy methods to build the query. As SQLAlchemy doesnot provide functions for soundex and levenshtein, we need to build those parts via RAW text. 

The words added in those texts are not direct user inputs, but processed ones. We take the sentences given by user, remove punctuation, normalize and split them into words(and romanized in two places) and these words are added to the text part of the query. 
Before it was added via python string formatting but now using parameters to the SQLAlchemy text function.

Also, this query is run as a SQLAlchmey query and not directly executed via their `exec` function. So I hope we have enough projection from injections here. Does it look okay to you, @joelthe1 ?

joelthe1:
LGTM 👍 Could you confirm that there are no other instances where are passing raw strings to SQL?

kavitharaju:
@joelthe1 .
Yes, I have already changed four other places `text()` was used in that same function. It is not used anywhere else.

Issue #230, opened 2021-08-31
New APIs for Auth


kavitharaju:
Pre-register the SuperAdmin User via the Kratos configs or App's initial setups. Credentials to be obtained from ENV variables.
New APIs needs to be added for:
- Registration: With Username, Password, email (to be verified), user app info(values: Ag, Vachan, None) that gives default roles to users 
- Login: only with username and password for now. Check how login with a 3rd party Auth service would work(how we can register a new user like this and how to identify same user upon next logins etc) 
- Logout 
- Add user role: Only Super admin can perform this for now, to create Vachan-Admin and AgAdmin. VachanUser and AgUser role creation should happen in registration process. Consider how to handle the case of “a Vachanuser now wants to become(register as) an AgUser too”.

PR #232, opened 2021-09-02
Kratos auth restapi

sijumoncy:
Primary implementation of Authentication of rest Api based on Ory Kratos. Email verification of registration is currently not added in this PR

create_super_user() is commented in main.py which cause error in github and it should uncomment after setting kratos server locally on test

#230


app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+#No name 'BaseModel' in module 'pydantic' (no-name-in-module)
+#pylint: disable=E0611
+from pydantic import BaseModel
+
+#Too few public methods (0/2) (too-few-public-methods)
+#pylint: disable=R0903
+class AuthDetails(BaseModel):
+    """schema for authentication input"""
+    username:str
+    password:str
```
kavitharaju, line 11: passwords shouldn't be sent as plain text. There would be data type used for passwords. Use it even if we are changing this to query params

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+
+#pylint: disable=R0903
+class Registration(BaseModel):
+    """kratos registration input"""
+    email:str
+    password:str
```
kavitharaju, line 17: dont use str for password here also

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+    firstname:str
+    lastname:str
```
kavitharaju, line 19: This field can be optional, I think

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+    appname:str
```
kavitharaju, line 20: Define an enum for valid app names and use that as the type instead on any str. This filed can also be optional

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+
+#pylint: disable=R0903
+class UserRole(BaseModel):
+    """kratos user role input"""
+    userid:str
+    roles:list
```
kavitharaju, line 26: We can have an enum for valid roles names as well. Define these enums in one place(either here or in the authentication.py file) and import and use the same every where instead of using strings

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+class UserIdentity(BaseModel):
+    """kratos user role input"""
+    userid:str
+
+#pylint: disable=R0903
+class RegistrationOut(BaseModel):
+    """registration output"""
+    id:str
+    email:str
+    Permisions:list
```
kavitharaju, line 38: specify what kind of list this is. I think we can specify types as `[str]` , `[int]` etc. If these permissions are user roles use the same enum as before

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+
+#pylint: disable=R0903
+class RegisterResponse(BaseModel):
+    """Response object of registration"""
+    details:str
```
kavitharaju, line 43: If these `details` field is like the "message" we use in other responses, it is better to rename them to "message" to have uniformity.

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+#pylint: disable=R0903
+class LogoutResponse(BaseModel):
+    """Response object of logout"""
+    message:str
+
+#pylint: disable=R0903
+class CommmonError(BaseModel):
+    """login error"""
+    error:str
+    details:str
```
kavitharaju, line 62: I think we already have a schema defined for error.

app/schema_auth.py
```diff
@@ -0,0 +1,74 @@
...
+
+#pylint: disable=R0903
+class UseroleResponse(BaseModel):
+    """user role update response"""
+    details:str
+    role_list:list
```
kavitharaju, line 68: define enum and use `[enum_name]` as type

kavitharaju (changes requested):
This is a lot of code to review. 
I haven't reviewed tests or tested locally yet. 
I have made sooo many comments already. Lets get through them first.

app/authentication.py
```diff
@@ -0,0 +1,314 @@
...
+from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
+#pylint: disable=E0401
+#pylint gives import error if not relative import is used. But app(uvicorn) doesn't accept it
+from custom_exceptions import GenericException , PermisionException ,\
+     AlreadyExistsException,NotAvailableException
+
+PUBLIC_BASE_URL = os.environ.get("KRATOS_PUBLIC_BASE_URL", "http://127.0.0.1:4433/self-service/")
+ADMIN_BASE_URL = os.environ.get("KRATOS_ADMIN_BASE_URL", "http://127.0.0.1:4434/")
+USER_SESSION_URL = os.environ.get("KRATOS_USER_SESSION_URL",\
+     "http://127.0.0.1:4433/sessions/whoami")
```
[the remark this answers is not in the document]
sijumoncy: For public and admin the ports are different and base is same and also the 3 rd one can be created from the admin iits self

app/schema_auth.py
sijumoncy: I think all field in the kratos schema should provide

app/authentication.py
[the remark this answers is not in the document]
kavitharaju: okay, then keep as two different values

kavitharaju (changes requested):
Suggestions
1. start env variable names with VACHAN_
2. In the error response, the "error" should say "Authentication Error" intead of "Http Error". Define a new custom exception for this if required. In "details" its better to give what ever error message you get from Kratos intead of giving our own texts like "Invalid Credential".
3. In the login api, use "user_email" instead of "username" as we ask for first name and last name also in registaration, this can be confusing
4. Please change the order of including router for Auth_apis in main.py, so that Authentication APIs willcome at the top of API docs and not at the end
5. Add new folder named Kratos_config. Add the config files needed there and also write an instructions.md file with the steps and commands we need to start the Kratos instance locally. This is for temporory purpose. Kratos will be setup separately. But noting down our methods now can help us work smoothly now and later to set up the central instance.
6. Should do logging in auth APIs(routers) also, like we have for other APIs

7. I faced following problems when testing locally
	1. Even though app started successfully(creating super user), was not able to login with super user credentials
	2. When I tried to register a new user with not safe password, the actual Kratos response was "'The password can not be used because the password has been found in data breaches...", but the API just gave a null with 201 status. I think the issue is at the check `if "messages" in reg_response["ui"]`. In this case `reg_response["ui"]` is an array.
	3. Throws Keyerror: 'userrole' at line 170, authentication.py, when I used superuser email agian for registartion

sijumoncy:
#234 
Added ENV variable TEST_MODE in github action to "handle super user creation" cause error because of the absence of kratos instance

kavitharaju (approved):
The some of the auth tests are still failing for me.

One reason for test failure is that the users created once are persisting in the Kratos DB and whe run again, it gives "Already Exists".

But merging this PR as other features are working. Tests can be fixed as another issue

Issue #235, opened 2021-09-08
Code quality improvement


kavitharaju:
In V2 we are using pylint to ensure code quality. 
In the first phase of development we have disabled some of their rules, because the aim was to get a working system ready.

Now, need to fix those issues as per the recommended standards.

PR #239, opened 2021-09-21
Linitng v2

sijumoncy:
#235 
Pylint quality improvement for v2

### DIFF
--- a/app/.pylintrc
+++ b/app/.pylintrc
@@ -1,0 +1,633 @@
+[MASTER]
+init-hook='from pylint.config import find_pylintrc; import os, sys; sys.path.append(os.path.dirname(find_pylintrc()))'
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-allow-list='pydantic'
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
+# for backward compatibility.)
+extension-pkg-whitelist=
+
+# Return non-zero exit code if any of these messages/categories are detected,
+# even if score is above --fail-under value. Syntax same as enable. Messages
+# specified are enabled, while categories only check already-enabled messages.
+fail-on=
+
+# Specify a score threshold to be exceeded before program exits with error.
+fail-under=10.0
+
+# Files or directories to be skipped. They should be base names, not paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the ignore-list. The
+# regex matches against paths.
+ignore-paths=
+
+# Files or directories matching the regex patterns are skipped. The regex
+# matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=yes
+
+
+[MESSAGES CONTROL]
+
+# Only show warnings with the listed confidence levels. Leave empty to show
+# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED.
+confidence=
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once). You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use "--disable=all --enable=classes
+# --disable=W".
+disable=print-statement,
+        parameter-unpacking,
+        unpacking-in-except,
+        old-raise-syntax,
+        backtick,
+        long-suffix,
+        old-ne-operator,
+        old-octal-literal,
+        import-star-module-level,
+        non-ascii-bytes-literal,
+        raw-checker-failed,
+        bad-inline-option,
+        locally-disabled,
+        file-ignored,
+        suppressed-message,
+        useless-suppression,
+        deprecated-pragma,
+        use-symbolic-message-instead,
+        apply-builtin,
+        basestring-builtin,
+        buffer-builtin,
+        cmp-builtin,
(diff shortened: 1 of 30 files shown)


app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once). You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use "--disable=all --enable=classes
+# --disable=W".
+disable=print-statement,
+        parameter-unpacking,
+        unpacking-in-except,
```
kavitharaju, line 80: why are we disabling such a long list of rules?

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+# Naming style matching correct class attribute names.
+class-attribute-naming-style=any
+
+# Regular expression matching correct class attribute names. Overrides class-
+# attribute-naming-style.
+#class-attribute-rgx=
+
+# Naming style matching correct class constant names.
+#class-const-naming-style=UPPER_CASE
+class-const-naming-style=any
```
kavitharaju, line 240: was this changed by us to support any specific case?

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+
+# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
+# tab).
+indent-string='    '
+
+# Maximum number of characters on a single line.
+max-line-length=100
+
+# Maximum number of lines in a module.
+max-module-lines=1400
```
kavitharaju, line 348: If you changed this, I think we should stick to the default 1000 and consider modularizing our larger files

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+
+
+[VARIABLES]
+
+# List of additional names supposed to be defined in builtins. Remember that
+# you should avoid defining new builtins when possible.
+additional-builtins=
+
+# Tells whether unused global variables should be treated as a violation.
+allow-global-unused-variables=yes
```
kavitharaju, line 496: is this a detault setting? or did we change it?

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+
+# Maximum number of attributes for a class (see R0902).
+max-attributes=7
+
+# Maximum number of boolean expressions in an if statement (see R0916).
+max-bool-expr=5
+
+# Maximum number of branch for function / method body.
+#max-branches=12
+max-branches=30
```
kavitharaju, line 562: Did we change it? I think if we are making exceptions and really cn't change the function logic then it is better to disable the rule at one place in code rather than making huge exceptions like this gloablly

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+
+# Maximum number of locals for function / method body.
+#max-locals=15
+max-locals=35
```
kavitharaju, line 566: Here also, if we really cannot re-write some functions, disable the rule there only. Dont make such huge changes globally.

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+
+# Maximum number of parents for a class (see R0901).
+max-parents=7
+
+# Maximum number of public methods for a class (see R0904).
+#max-public-methods=20
+max-public-methods=30
```
kavitharaju, line 573: same as max-locals and max-branches

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+
+# Maximum number of return / yield for function / method body.
+max-returns=6
+
+# Maximum number of statements in function / method body.
+#max-statements=50
+max-statements=65
```
kavitharaju, line 580: Same as above comments

app/.pylintrc
```diff
@@ -0,0 +1,633 @@
...
+
+# Minimum number of public methods for a class (see R0903).
+#min-public-methods=2
+min-public-methods=0
```
kavitharaju, line 584: Same as above comments

kavitharaju (commented):
I have some concerns about disabling and tweaking rules globally. I am afraid it may affect the quality of the code thorough out. The reason we want to use pylint is, because we want to adhere to their rules as much as we can.

app/.pylintrc
sijumoncy: In our case the enums are not defined as class constants in upper case . We used snake case throughout the code

app/.pylintrc
sijumoncy: Default . All our changes have a comment with original values

app/.pylintrc
sijumoncy: Wr have this case throughout crud moduke

app/.pylintrc
kavitharaju: If using UPPER_CASE is the recommended standard we should try to change our code that way

app/.pylintrc
kavitharaju: As discussed offline, we need to try and refactor our code and follow the recommended standard as much as possible

app/crud/nlp_crud_utils.py
```diff
@@ -0,0 +1,284 @@
+''' Place to define all utitlity functions
```
kavitharaju, line 1: How about a different name for the file which will let us know what is in there?
One suggestion, `nlp_tokenization.py` , that is if ALL tokenization related things are moved here.

app/crud/projects_crud.py
```diff
@@ -5,17 +5,12 @@
 from sqlalchemy.orm import Session
 from sqlalchemy.orm.attributes import flag_modified
 
-#pylint: disable=E0401
-#pylint gives import error if not relative import is used. But app(uvicorn) doesn't accept it
-
```
kavitharaju, line 10: How is the issue of relative import fixed? Is it by disabling it from the rc file?

app/graphql_api/mutations.py
```diff
@@ -1,5 +1,5 @@
 '''GraphQL queries and mutations'''
-#pylint: disable=C0302
+#pylint: disable=C0302,C0103
```
kavitharaju, line 2: is c0103 related to snake casing function names? Are we disabling it because of some ghraphene convention/requirement? If yes, one possibility is that Graphene will convert the casing we used to its required format like it does for field names. If not, please check if we have options to whitelist graphene like we do for pydantic

kavitharaju (commented):
The way you have split some functions looks good to me. 
But if this PR fixes only the issues in some files, then please remove the the other files from the PR which may require more work.

app/crud/nlp_crud_utils.py
sijumoncy: I moved some function to a new file to reduce the total length of file .
All functions that i moved are nlp utility functions .

app/crud/projects_crud.py
sijumoncy: Yes there is a hook added in rcfile at top 

Which help pylint to identify the working directory and modules

app/graphql_api/mutations.py
sijumoncy: THIS PR only contain the fixes related to CRUD MODULE only . We wil add the remaining fixes in the next pr . C0103 is the issue for constants declaration of class in capital letter

app/crud/nlp_crud_utils.py
kavitharaju: I wouldn't call them "utility" functions. They some core NLP operations and logics we have. Hence I suggested renaming. And I see these are not really DB CRUD(create, read, update and delete) operations. So need not have crud also in the name even though we are placing them within the CRUD module.

I would prefer some name with "nlp" in it and that can summarize the contents in the best possible way

app/graphql_api/mutations.py
kavitharaju: That is why I am suggesting f removing those files that are not changed properly from this PR. On top of his page, you can see 29 files changed. Limit the files to only the ones in CRUD module

kavitharaju (approved):

Issue #231, opened 2021-08-31
 Show admin and translator all their organisations


Bobbykumar706584:
Currently, the only verified organization is showing for admin users and the translator can't see anything. 
So If a user is not super admin we can show all their organization without putting verified = true check.
URL = "/v1/autographamt/organisations", methods=["GET"]

RevantCI:
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/12570574/131481257-97c08167-ca3e-4ce1-b345-d704b6129b9f.png)
Adding screenshot of UI for organization view for admin that is being made, the translator will also have a similar view the right menu may be different for him.
Currently only verified organizations can be fetched as seen in the sceenshot.
This is to help the user(admin/translator) to see the status of the pending organizations for which the request(s) have been sent.

PR #254, opened 2021-09-29
organization requests  #231

RevantCI:
1. #231  
2. Typo fixed

### DIFF
--- a/agmt/main.py
+++ b/agmt/main.py
@@ -349,15 +349,13 @@ def autographamtOrganisations():
 				organisation_phone, organisation_email, verified, user_id, status from autographamt_organisations\
 					order by organisation_id")
 			rst = cursor.fetchall()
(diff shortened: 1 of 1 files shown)


kavitharaju (approved):
The changes looks okay to me, if the requirement is just to show "unverified" organizations also on UI. 
But I noticed that in the requirement issue it says show admins and _translators_ their organizations. And I also see that you have changed a line in code `elif role==2`(admin) to `else`(both admins and translators). But as per our implementation we dont have a concept of translators belonging to an org. Also, as far as you are using `user_id` in query, it is doing an admin only fetch. If there any particluar case that will be covered here like org request send and user is still a translator which requires this. Did you check via UI.
This doesn't look like a problem to me. Just wanted to point it out.

Bobbykumar706584 (approved):

Issue #256, opened 2021-10-01
Suggestion on access checking implementation


kavitharaju:
some thoughts and suggestions on the current code

1. instantiation of auth handler is done in auth_api.py. `auth_handler = AuthHandler()`. This is to be used in all routers for verifying token, right? Should it be done in the authentication.py  and shouldn't the onject be imported instead of the class?
2. `auth_handler.kratos_session_validation()` returns only user roles. In some cases we would need user_id also to check resource specific access. So that should also be returned.
3. The access checking code snippet, that is to be repeated in all APIs should be shortened. Consider the possibility of using a decorator.
```
    request_context = get_request_context(request)

    #TEst for new access right function
    resource_id =None
    verified = check_access_rights(db_, resource_id, request_context,
        user_id=None, user_roles=None,resource_type = None)
    if verified:
        data = user_register_kratos(register_details,app_type)
    else:
        raise PermisionException("Access Permission Denied for the URL")
```
The `resource_id` values we need to check would be available only post function execution. Also we may sometimes have to raise exception(eg. when a non-owner user trying to edit project), somethimes we may just have to filter out some items from the return list(eg. when user fetching available commentaries from vachan-online, via get sources API). Consider the possibility of doing all this from a single decorator function.
4. Use variable name `user_roles` here not `permissions` in
```
@router...()
def func_name(...., permision = Depends(auth_handler.kratos_session_validation),...):
	...
	...
	verified = check_access_rights(db_, resource_id, request_context,
        user_id=None, user_roles=permision,resource_type = None)

```
As per the suggestion in point 2, we may change it to `user_details = {user_id:"", user_roles:""}`

Issue #245, opened 2021-09-25
Auth and access checks in meta-contents APIs


kavitharaju:
Add required Auth and access checks to languages, contents, licenses and versions APIs. Add/update tests for them checking for access related cases too.

PR #257, opened 2021-10-06
V2 role based access

sijumoncy:
#245 #256 

This PR contains the Access Rule check implementation logic as per the last discussion for meta-contents. In the test cases  some errors are coming for graphql endpoints because of the db related changes are done in the meta content crud (commit and refresh). The other some errors are because of the kratos related user registration calls for the test cases (only in git test).i think this PR will make some sense about the implementation structure of access rule check according to the discussions.


app/test/test_languages.py
```diff
@@ -157,10 +200,77 @@ def test_post_missingvalue_language():
...
+    headers_auth2 = {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token2
+            }
+    response = client.put(UNIT_URL, headers=headers_auth2, json=data)
+    assert response.status_code == 403
+    assert response.json()['error'] == "Permision Denied"
+
+    #delete the user
+    delete_user_identity(test_user_id2)
```
kavitharaju, line 272: creating users and deleting  them needs to be done for all tests right? how about create a set of users in different roles at the start of testing process and delete them once all tests are run. I am thinking of using a fixture with scope session.

app/test/test_languages.py
```diff
@@ -2,9 +2,18 @@
...
 UNIT_URL = '/v2/languages'
 
+#create a normal user for this module test
+test_user_data = {
+        "email": "abc@gmail.com",
+        "password": "passwordabc@1"
+    }
+response = register(test_user_data,apptype='API-user')
+test_user_id = [response.json()["registered_details"]["id"]]
+test_user_token = response.json()["token"]
```
kavitharaju, line 16: I see you have added test with a nd without auth for post and edit tests. I would suggest doing it at least once for a get also. Even though auth is not required for it, need to make sure API will function even when auth header is provided.

kavitharaju (commented):
The access check steps look good to me at this point. Lets move on with the implementation and revise our approaches later if required

app/test/test_languages.py
kavitharaju: okay. Shall add it as separate issue and we can do it in later PRs

Issue #258, opened 2021-10-07
Pytest fixture for test users


kavitharaju:
For running tests, we need to create and delete test users. This is now done by creating new users in all required test functions and deleting them at the end of that test function.
 This could be done as a fixture that runs just once, creating a bunch of test users with different roles at the start of a test session and deleting them all at the end of the session. This way the number of times users are created and deleted from Kratos will be reduced considerably

Issue #247, opened 2021-09-25
Auth and access check for sources APIs


kavitharaju:
Add auth and access check for sources APIs.

Add tests. Some possible tests are, check if the list of sources fetched for different users are different. For example, a VachanUser accessing `GET v2/sources` should see only the items with `open-access` permission, but a `VachanAdmin` user should see all the avialable contents. Another type of tests are to ensure only a user who created an item or `SuperAdmin` can edit its fields or metaData

Issue #246, opened 2021-09-25
Mention permission of sources


kavitharaju:
Add provision to specify permissions like "open-access", "publishable", "downloadable", "derivable" to `sources` in POST and PUT APIs. These are to be stored in the metaData field in the table under `"access-permissions"` key as a list of values. 

By default all sources should have a "content" permission added.

This is expected to be used from the Vachan-admin app when contents are uploaded to or maintained in the server.

PR #259, opened 2021-10-13
V2 role based access

sijumoncy:
#246 #247 #258 

This PR contains the Authentication and access rule check related with Role Based Access System implementation. As we discussed the things are implemented. The test cases are added separately for POST PUT and GET with different possibilities and get is also filtered with roles and apps. Some issues are caused in the other tests due to the auth logic and all auth implemented end points (Meta Content and source) test are working properly.

### DIFF
--- a/app/authentication.py
+++ b/app/authentication.py
@@ -3,8 +3,9 @@ import os
 import json
 from functools import wraps
 import requests
-from fastapi import HTTPException, Security
-from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
+from fastapi import HTTPException, Security , Depends
+from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer ,\
+    OAuth2PasswordBearer
 from sqlalchemy.orm import Session
 import db_models
 import schema_auth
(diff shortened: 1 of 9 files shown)


app/routers/content_apis.py
```diff
@@ -12,6 +13,13 @@
 router = APIRouter()
 auth_handler = AuthHandler()
 
+#optional authentication with token or none
+optional_oauth2_scheme = OAuth2PasswordBearer(tokenUrl="auth", auto_error=False)
+async def get_user_or_none(token: str = Depends(optional_oauth2_scheme)):
+    """optional auth for getting token of logined user not raise error if no auth"""
+    # print("token===>",token)
```
kavitharaju, line 20: I assume this is creating a dummy user token with None values, in the absence of actual token in the header. This looks like a good way of handling the `noauthrequired` cases. But a few thoughts about it.

As other endpoints defined else where also might need this, may be we can move this definition to a common place like `authentication.py` and import to different sub modules in routers is required.

Another thought is this. Even though we are specifying different access rights associated with different endpoints in the  `api_permission_map.py` the authetication part is now specified at the endpoint definition in the router module itself. We have 3 ways we do it now. 
1. For endpoints that dont need an auth, we omit the Dependancy
2. For those that need an auth we use  `Depends(auth_handler.kratos_session_validation)`
3. For those end points that has optional auth requirement we use, `Depends(get_user_or_none)`
Can combine all these into one function and call that same function uniformly as dependancy of every endpoint? Let it be this `get_user_or_none()` function. In the access check decorator,  we can raise error or accept the None user, based on the APIs permission map. That way all the endpoint specific auth specifying will be required only at that one file.

app/crud/structurals_crud.py
```diff
@@ -291,7 +291,7 @@ def create_source(db_: Session, source: schemas.SourceCreate, source_name, user_
         metaData = source.metaData,
```
kavitharaju, line 291: Doesn't this part(both in create source and update source methods) require modification to store the access tags specified by content uploader in the DB?

app/authentication.py
```diff
@@ -283,26 +326,36 @@ async def wrapper(*args, **kwargs):
...
-                    raise PermisionException("Access Permission Denied for the URL")
-                db_.commit()#pylint: disable=E1101
-                db_.refresh(db_resource)#pylint: disable=E1101
-            elif required_params['request_context']['method'] == 'GET':#pylint: disable=E1126
-                db_resource = response
-                verified = check_access_rights(db_, required_params, db_resource)
-                if not verified:
-                    raise PermisionException("Access Permission Denied for the URL")
+            if len(response) > 0:
+                if required_params['request_context']['method'] != 'GET':#pylint: disable=E1126
```
kavitharaju, line 339: I assume this case is added for update or create functions. In that case, when the response in first obtained in this decorator before the `db.commit()`, please make sure the final response include `id` values also.

sijumoncy: fixed id is coming

app/routers/content_apis.py
sijumoncy: work on this idea to combine auth function

Issue #260, opened 2021-10-22
Change sources APIs with access tags


kavitharaju:
* In GET `/v2/sources`, add an optional query param to filter contents based on access tag given for them.
* In PUT `/v2/sources`, add provision to update(add/remove) access tags for a particular source table

PR #263, opened 2021-10-29
V2 role based access

kavitharaju:
#260 
In GET /v2/sources, add an optional query param to filter contents based on access tag given for them.
PR contains the filter option for get source based on the provided access tags.

In PUT /v2/sources, add provision to update(add/remove) access tags for a particular source table
     This is already covered in the previous PR for sources. PUT for v2/sources have an option to update the access tags for the source table

### DIFF
--- a/app/test/test_sources.py
+++ b/app/test/test_sources.py
@@ -404,13 +404,7 @@ def test_created_user_can_only_edit():
         "sourceName": 'ml_TTT_2_commentary',
         "revision": 1
     }
-    # vachanAdmin_user_data = {
-    #         "user_email": "vachanadmintest@mail.test",
-    #         "password": "passwordtest@1"
-    #     }
-    # response = login(vachanAdmin_user_data)
-    # assert response.json()['message'] == "Login Succesfull"
-    # test_user_token = response.json()["token"]
+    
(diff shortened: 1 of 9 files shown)


sijumoncy:
Bug Fixes updated : #264 #265 

Test failing in gql project db
Kratos Url bug

app/test/test_sources.py
```diff
@@ -584,6 +568,69 @@ def test_get_after_adding_data():
...
+    data['language'] = 'mr'
+    data['accessPermissions'] = ['open-access']
+    check_post(data)
+    data['language'] = 'te'
+    data['accessPermissions'] = ['publishable']
+    check_post(data)
+    
+    response1 = client.get(UNIT_URL,headers=headers_auth)
+    assert response1.status_code == 200
+    assert len(response1.json()) == 3
```
kavitharaju, line 597: I recommend using a version filter also here and others similar requests below.
If we run these tests on a DB with already existing contents, there might be difference in number of items returned. So query for only the "TTT" contents we add via this test.

kavitharaju (commented):
Only one suggestion for now

app/test/test_sources.py
sijumoncy: ok

kavitharaju (commented):
Approved!

PR #276, opened 2021-11-06
Feature/dockerization

kavitharaju:
Fix issues in the dockerization.


kavitharaju:
@RevantCI  Thanks for checking and the suggestions.
As it is found to be working for Siju also on windows and ubuntu now, merging it as such.

Issue #250, opened 2021-09-25
Auth and access check for Autographa Project APIs


kavitharaju:
Add auth and access check for autographa project management as well as translation APIs. 

Add tests, that check things like get projects will list user specific projects(only the ones where requesting user is owner or member). Who all can tokenize, generate suggestions, confirm tranlsation or view drafts in a proect. Can an internal user(BcsDeveloper) get list of and access drafts of all projects?

PR #280, opened 2021-11-12
V2 role based access

sijumoncy:
#250 
This PR contains only auth and access check for autographa project management and not for translation APIs.


app/authentication.py
```diff
@@ -279,26 +280,65 @@ def filter_resource_content_get(db_resource, access_tags, required_permission, u
...
+        if len(filtered_content) == 0:
+            for project in db_resource:
+                project_user_obj = project.users
+                for user in project_user_obj:
+                    if user_id == user.userId and\
+                        user.userRole in allowed_users and \
+                            not any(project == dic for dic in filtered_content):
+                        filtered_content.append(project)
+
+    if len(filtered_content) > 0:
```
kavitharaju, line 313: What happen if this API is called to fill the home page of a  new user who doesn't have any projects? Will this throw an error?

app/authentication.py
```diff
@@ -478,15 +527,22 @@ def kratos_logout(self,auth:HTTPAuthorizationCredentials= Security(security)):
         return data
 
 #get all user details
-def get_all_kratos_users():
-    """get all user info"""
+def get_all_kratos_users(rec_user_id=None):
```
kavitharaju, line 530: better to remove "all" from the function name as it is used to get either all or a specific user now.

sijumoncy: ok

app/authentication.py
sijumoncy: yes currently it will throw an error only .But we can change it to [ ]

app/test/test_agmt_projects.py
```diff
@@ -775,20 +775,17 @@ def test_get_project_access_rules():
...
     headers_auth['Authorization'] = "Bearer"+" "+initial_test_users['APIUser']['token']
     response = client.get(UNIT_URL,headers=headers_auth)
-    assert response.status_code == 403
-    assert response.json()['error'] == 'Permision Denied'
+    assert len(response.json()) == 0
     headers_auth['Authorization'] = "Bearer"+" "+initial_test_users['VachanUser']['token']
     response = client.get(UNIT_URL,headers=headers_auth)
-    assert response.status_code == 403
-    assert response.json()['error'] == 'Permision Denied'
+    assert len(response.json()) == 0
```
kavitharaju, line 788: There is a check for notAvailableContent you could use here. Just letting you know for future cases.

kavitharaju (approved):

PR #281, opened 2021-11-15
GET API for sw

anjalyv:
- V2/lookup/stopwords/<lang_code>
Included filters for specific word, system-defined/user-defined/auto-generated , active
- As input accepting language code. Skip and limit are also provided.
- Response will be like:
[
  {
    "stopword": "के",
    "stopwordType": "System defined",
    "confidence": null,
    "active": true
  }....
]
Confidence field will have score only if stopword is auto generated.


app/crud/nlp_sw_crud.py
```diff
@@ -0,0 +1,52 @@
...
+    query = query.offset(skip).limit(limit).all()
+    result = []
+    for row in query:
+        sw_type = ''
+        if row.confidence == 2:
+            sw_type = 'System defined'
+        elif row.confidence == 1:
+            sw_type = 'User defined'
+        else:
+            sw_type = 'Auto generated'
```
kavitharaju, line 46: Better use ENUM for these strings

app/db_models.py
```diff
@@ -115,14 +114,6 @@ def book_id(cls): # pylint: disable=E0213
...
         '''For modelling the book field in derived classes'''
         return relationship(BibleBook)
-    @hybrid_property
-    def ref_string(self):
-        '''To compose surrogate id'''
-        return f'{self.book.bookCode} {self.chapter}:{self.verseStart}-{self.verseEnd}'
-    @ref_string.expression
-    def ref_string(cls): # pylint: disable=E0213
-        '''To compose surrogate id'''
-        return func.concat(BibleBook.bookCode," ",cls.chapter,":",cls.verseStart,"-",cls.verseEnd)
```
kavitharaju, line 125: why are these code lines removed?

app/db_models.py
```diff
@@ -229,15 +220,6 @@ def book_id(cls): # pylint: disable=E0213
...
         return relationship(BibleBook)
-    @hybrid_property
-    def ref_string(self):
-        '''To compose surrogate id'''
-        return f'{self.book.bookCode} {self.chapter}:{self.verseNumber}'
-
-    @ref_string.expression
-    def ref_string(cls): # pylint: disable=E0213
-        '''To compose surrogate id'''
-        return func.concat(BibleBook.bookCode," ",cls.chapter,":",cls.verseNumber)
```
kavitharaju, line 240: Why are these also removed?

kavitharaju (changes requested):
1. I think your PR doesnot include latest changes from the version-2 branch. Some code lines are shown as being removed. I suggest you merge version2 to this branch
2. There is linting error shown
3. We need to add tests for all the new APIs we implement

PR #285, opened 2021-11-17
Get api to retrieve stop words

anjalyv:
/v2/lookup/stopwords/<lang_code>
* Included filters for specific word, system-defined/user-defined/auto-generated , active
* As input accepting language code. Skip and limit are also provided.
Response will be like:
[
{
"stopword": "के",
"stopwordType": "System defined",
"confidence": null,
"active": true
}....
]
Confidence field will have score only if stopword is auto generated.


app/crud/nlp_sw_crud.py
```diff
@@ -0,0 +1,53 @@
...
+        if row.confidence == 2:
+            sw_type = schemas_nlp.StopWordsType.SYSTEM.value
+        elif row.confidence == 1:
+            sw_type = schemas_nlp.StopWordsType.USER.value
+        else:
+            sw_type = schemas_nlp.StopWordsType.AUTO.value
+        conf_val = None
+        if row.confidence not in [1, 2]:
+            conf_val = row.confidence
+        result.append({"stopword": row.stopWord, "stopwordType": sw_type, "confidence": conf_val,
```
kavitharaju, line 51: Dont we have a metadata field too? That should be returned too.

app/test/test_stop_words_generation.py
```diff
@@ -0,0 +1,56 @@
...
+
+UNIT_URL = '/v2/lookup/stopwords'
+headers = {"contentType": "application/json", "accept": "application/json"}
+
+def assert_positive_get_stopwords(item):
+    '''Check for the properties in the normal return object'''
+    assert "stopword" in item
+    assert "stopwordType" in item
+    assert "confidence" in item
+    assert "active" in item
```
kavitharaju, line 15: Include check for metadata also as optional field. That is, if present check for its type.

app/test/test_stop_words_generation.py
```diff
@@ -0,0 +1,56 @@
...
+
+def test_get_stop_words():
+    '''Positve tests for get stopwords API'''
+    default_response = client.get(UNIT_URL+'/hi?', headers=headers) 
+    assert default_response.status_code == 200
+    assert isinstance(default_response.json(), list)
+    for item in default_response.json():
+        assert_positive_get_stopwords(item)
+        if item["stopwordType"] in ["system defined", "user defined"]:
+            assert item["confidence"] is None
```
kavitharaju, line 29: Are we returning a None for confidence. That field should be eliminated from the response if its value is None. Use the parameter response_model_exclude_none in router definition to do that

app/crud/nlp_sw_crud.py
```diff
@@ -35,19 +34,9 @@ def retrieve_stopwords(db_: Session, language_code, **kwargs):
...
-        elif row.confidence == 1:
-            sw_type = schemas_nlp.StopWordsType.USER.value
-        else:
-            sw_type = schemas_nlp.StopWordsType.AUTO.value
-        conf_val = None
-        if row.confidence not in [1, 2]:
-            conf_val = row.confidence
-        result.append({"stopword": row.stopWord, "stopwordType": sw_type, "confidence": conf_val,
-        "active": row.active})
+    for row in query_result:
```
kavitharaju, line 39: Is it required to convert the SQLAlchemy object `query_result` to a dict? Can't we return the `query_result` as such?

app/schemas_nlp.py
```diff
@@ -235,17 +235,30 @@ class Translation(BaseModel):
...
+    '''Types of stop-words based on how they are generated'''
+    SYSTEM = 'system defined'
+    USER = 'user defined'
+    AUTO = 'auto generated'
+
 class StopWords(BaseModel):
     '''Response object for stop words'''
     stopword: str = Field(..., example="और")
-    stopwordType: str = Field(..., example="Auto generated")
+    stopwordType: str = Field(None, example="Auto generated")
```
kavitharaju, line 247: Why is `stopwordType` made optional(None)? Also why is its type `str`, shouldn't it be the enum `StopWordsType`?

kavitharaju: Specifying it as the enum will give proper API docs for the users as to waht values to expect there

app/crud/nlp_sw_crud.py
anjalyv: While returning `query_result` ie, db_models object, getting validation-error saying that value is not a valid dict. Few lines of error :

`pydantic.error_wrappers.ValidationError: 94 validation errors for StopWords
response -> 0
  value is not a valid dict (type=type_error.dict)`

kavitharaju (approved):

PR #288, opened 2021-11-22
Put api to update stopword fields in db

anjalyv:
- url : v2/lookup/stopwords/<lang_code>

- To update metadata or active status

- Input should specify a single sw to be edited
Eg: {
  "stopWord": "कहता",
  "active": false,
  "metaData": {
    "type": "verb"
  }
}

- Altered rows will be returned as response including status message

### DIFF
--- a/app/crud/nlp_sw_crud.py
+++ b/app/crud/nlp_sw_crud.py
@@ -49,10 +49,13 @@ def update_stopword_info(db_: Session, language_code, sw_json):
         raise NotAvailableException("Language with code %s, not in database"%language_code)
     language_id = language_id[0]
     stopword = sw_json.stopWord
-    active_val = sw_json.active
-    metadata = sw_json.metaData
+    value_dic = {}
+    if sw_json.active is not None:
+        value_dic["active"] = sw_json.active
+    if sw_json.metaData is not None:
+        value_dic["metaData"] = sw_json.metaData   
     update_stmt = (update(db_models.StopWords).where(db_models.StopWords.stopWord == stopword,
-        db_models.StopWords.languageId == language_id).values(active=active_val, metaData=metadata))
+        db_models.StopWords.languageId == language_id).values(value_dic))
     result = db_.execute(update_stmt)
     db_.commit()
     query = db_.query(db_models.StopWords)
--- a/app/crud/nlp_sw_crud.py
+++ b/app/crud/nlp_sw_crud.py
@@ -2,7 +2,7 @@
     to stop word identification'''
 
 from sqlalchemy.orm import Session
-from sqlalchemy import func
+from sqlalchemy import func, update
 from custom_exceptions import NotAvailableException
 
 import db_models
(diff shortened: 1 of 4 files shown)


app/schemas_nlp.py
```diff
@@ -262,3 +262,15 @@ def set_stopword_type(cls, values): # pylint: disable=R0201 disable=E0213
             if values['confidence'] in [1, 2]:
                 values['confidence'] = None
         return values
+
+class StopWordUpdate(BaseModel):
+    '''Import object for updating stopword info'''
+    stopWord: str = Field(..., example="और")
+    active : bool = Field(..., example=True)
```
kavitharaju, line 269: Active field should be optional(with None for default)

kavitharaju (changes requested):

app/test/test_stop_words_generation.py
```diff
@@ -74,16 +78,24 @@ def test_get_notavailable_code():
...
     for item in response.json()['data']:
         assert_positive_get_stopwords(item)
-    if response.json()['message'] == "Stopword info updated successfully":
-        assert len(response.json()['data']) == 1
+    assert response.json()['message'] == "Stopword info updated successfully"
+    assert len(response.json()['data']) == 1
 
-    response = client.put(UNIT_URL+'/hi?',headers=headers, json=wrong_obj)
+    response = client.put(UNIT_URL+'/hi?',headers=headers, json=update_wrong_obj)
     assert response.status_code == 200
```
kavitharaju, line 98: Are we not returning error code and error object here?

app/crud/nlp_sw_crud.py
```diff
@@ -40,3 +40,33 @@ def retrieve_stopwords(db_: Session, language_code, **kwargs):
...
+    rows = query.filter(db_models.StopWords.stopWord == stopword,
+            db_models.StopWords.languageId == language_id).all()
+    data = []
+    for row in rows:
+        data.append({"stopWord": row.stopWord, "confidence": row.confidence, "active": row.active,
+            "metaData": row.metaData})
+    if result.rowcount > 0:
+        msg = "Stopword info updated successfully"
+    else:
+        msg = "Couldn't update the provided data"
```
kavitharaju, line 71: What is this scenario? When provided lang_id and sw doesn't match any row in DB? If that is the caseThen I think we should return a 404 error(NotAvailableException).

app/test/test_stop_words_generation.py
```diff
@@ -51,3 +75,27 @@ def test_get_notavailable_code():
...
     response = client.get(UNIT_URL+"/abc")
     assert_not_available_content(response)
+
+def test_update_stopword():
+    '''Positve tests for update stopwords API'''
+    response = client.put(UNIT_URL+'/hi?',headers=headers, json=update_obj1)
+    assert response.status_code == 200
+    assert_positive_update_stopwords(response.json())
+    for item in response.json()['data']:
+        assert_positive_get_stopwords(item)
```
kavitharaju, line 85: As we are updating a single stopword the data object should be a single dict not a list. Or would there be any case where we would need a list to be returned

kavitharaju (approved):

PR #291, opened 2021-11-23
POST API to add new SWs for a language

anjalyv:
- v2/lookup/stopwords/<lang_code>
-  Accept a list of stopwords like,
[
{
    "stopWord": "क्योंकि",
    "metaData": {
      "type": "preposition"
    }
  },
  {
   "stopWord": "ने",
}
]

- For each word :
    If present in db and is predefined or added by a user, leave it as such
    If already present, and is auto generated update its confidence =  1, updated_user, active =True
    If not present add them to DB with confidence score =1 , active =True


app/crud/nlp_sw_crud.py
```diff
@@ -68,3 +69,53 @@ def update_stopword_info(db_: Session, language_code, sw_json):
...
+                                "active" : True,
+                                "updatedUser": user_id
+                              }
+                update_stmt = (update(db_models.StopWords).where(db_models.StopWords.stopWord ==
+                        word_dic.stopWord, db_models.StopWords.languageId == language_id).values
+                        (update_args))
+                result = db_.execute(update_stmt)
+                if result.rowcount == 1:
+                    row = db_.query(db_models.StopWords).filter(db_models.StopWords.stopWord ==
+                        word_dic.stopWord, db_models.StopWords.languageId == language_id).first()
```
kavitharaju, line 107: why do we need to execute a new query here? wouldn't the `result` object have this row and values?

app/routers/translation_apis.py
```diff
@@ -358,3 +358,15 @@ def update_stop_words(language_code:schemas.LangCodePattern=Path(...,example="hi
     log.debug('language_code:%s, sw_info:%s',language_code, sw_info)
     sw_data = nlp_sw_crud.update_stopword_info(db_, language_code, sw_info)
     return { "message": "Stopword info updated successfully", "data":sw_data }
+
+@router.post('/v2/lookup/stopwords/{language_code}',
+    response_model=schemas_nlp.StopWordsAddResponse, response_model_exclude_none=True,
+    status_code=200, tags=['Generic Translation'])
+def add_stopwords(language_code:schemas.LangCodePattern=Path(...,example="hi"),
+    stopwords_list:List[schemas_nlp.StopWordsAdd]=Body(...),
```
kavitharaju, line 366: May be we can just have `stopwords_list:List[str]=Body(..., example=....)`

app/schemas_nlp.py
```diff
@@ -274,3 +274,14 @@ class StopWordUpdateResponse(BaseModel):
     '''Response object after updating metadata or active status'''
     message: str = Field(..., example="Stopword info updated successfully")
     data: StopWords
+
+class StopWordsAdd(BaseModel):
+    '''Import object for adding new stopwords in db'''
+    stopWord: str = Field(..., example="और")
+    metaData: dict = Field(None, example={
```
kavitharaju, line 281: As per our last discussion we had decided to not include metdata field in this POST. It will create confusion as to which metadata to be persisted in the scenario where we have an autogenarated version of same SW, whose confidence value get updated by this call

app/schemas_nlp.py
```diff
@@ -274,3 +274,14 @@ class StopWordUpdateResponse(BaseModel):
...
+        "type":'postposition'})
+
+class StopWordsAddResponse(BaseModel):
+    '''Response object after adding new stopwords in db'''
+    message: str = Field(..., example="Stopwords saved successfully")
```
kavitharaju, line 286: The example could be changed as per the changed message where we also give the number

app/test/test_stop_words_generation.py
```diff
@@ -92,3 +104,14 @@ def test_update_stopword():
 
     response = client.put(UNIT_URL+'/hi?',headers=headers, json=update_wrong_obj)
     assert response.status_code == 404
+
+def test_add_stopword():
+    '''Positve tests for add stopwords API'''
+    response = client.post(UNIT_URL+'/hi?',headers=headers, json=add_obj)
```
kavitharaju, line 110: Are we covering enough cases where added stopwords are present/not present, have different confidences, different active values etc?

also I would suggest we use a different dummy language name and not Hindi for tests as this may cause problem with the actual hindi data we have in a deployed instance

kavitharaju (changes requested):

app/crud/nlp_sw_crud.py
anjalyv: No `result` object does not return rows.

kavitharaju (commented):
I suggest we add more tests to cover all different cases of how SW adding can happen.
For example we could follow some pattern like this,
1. Add some tests for predefned case
    choose lang: hi

    1. add a known sw and make sure it is not getting updated
     2. add a content word as sw(to make sure it will get updated)
2. test for user-defined case
    1. choose an unknown langauge and upload sws in it
    2. Then try adding an already added sw again
    3. Add new SW to the same language
3. test for auto-genarated case
      this cannot be done until you finish the other API too, but mmake sure all cases are covreed then

kavitharaju (approved):

PR #304, opened 2021-12-15
Auto generate stopwords api

anjalyv:
POST V2/translation/stopwords/generate

- Running the api as background task
- Input: lang_code, optionally sentenceList, flag use_server_data=True, optional param GL_lang_code
- Genarates stopwords using available server-data for the given language and sentenceList(if given)
- If available sentences count is <1000 not possible to generate stopwords
- If GL is given, filters translation_words of GL from the generated stopwords
- Stores a generated stopword in db, if that word is not present in db
- If stop word already in db, updates its confidence, updated_user only if confidence in db is less than that of generated score
- Return the dynamically obtained SW list


app/test/test_stop_words_generation.py
```diff
@@ -85,13 +110,13 @@ def test_update_stopword():
     '''Positve tests for update stopwords API'''
     response = client.put(UNIT_URL+'/hi?',headers=headers, json=update_obj1)
     assert response.status_code == 200
-    assert_positive_update_stopwords(response.json())
+    assert_positive_response(response.json())
```
kavitharaju, line 113: shouldn't this be 
```
assert_positive_response(response.json()['data`])
```
?

app/test/test_stop_words_generation.py
```diff
@@ -128,15 +153,132 @@ def test_add_stopword():
...
+    "revision": 1,
+    "year": 2020,
+    "license": "CC-BY-SA",
+    "metaData": {"owner": "someone", "access-key": "123xyz"}
+    }
+    source = add_source(src_data)
+    table_name = source.json()['data']['sourceName']
+    data = []
+    for book in os.listdir('test/resources/'):
+        if book.endswith('usfm'):
```
kavitharaju, line 216: Wouldn't this add all usfm files in the resources folder? What if we add a new incorrect or different language file for another testt there? I suggest listing just the required files for this test

app/test/test_stop_words_generation.py
```diff
@@ -128,15 +153,132 @@ def test_add_stopword():
...
+            book_data = open('test/resources/' + book, 'r').read()
+            data.append({"USFM":book_data})
+    client.post('/v2/bibles/'+table_name+'/books', headers=headers, json=data)
```
kavitharaju, line 219: It is good to do some assertions after these kind of requests too to make sure things are working as expected

app/crud/nlp_sw_crud.py
```diff
@@ -113,3 +121,208 @@ def add_stopwords(db_: Session, language_code, stopwords_list, user_id=None):
     db_.commit()
     msg = f"{len(db_content)} stopwords added successfully"
     return msg, db_content
+
+def clean_text(text):
+    '''Cleaning text by removing punctuations, extra spaces'''
+    text = re.sub('[“”]', '"', text)
+    text = re.sub('[‘’]', "'", text)
+    text = re.sub('।', '', text)
+    text = text.translate(str.maketrans('', '', string.punctuation))
```
kavitharaju, line 130: Do we need to retain the quotes in text for the SW identification?
If not, there is a list of punctuation we have [defined in utils](https://github.com/Bridgeconn/vachan-api/blob/3485f010badf54aa50a1277f7d4e7f7dccfaefc0/app/crud/utils.py#L14) which can be used to replace all of them with space or something. 
It would be reuse the same set for all the operations.

app/crud/nlp_sw_crud.py
```diff
@@ -113,3 +121,208 @@ def add_stopwords(db_: Session, language_code, stopwords_list, user_id=None):
...
+        sw_scored.append((word, score))
+    return sw_scored
+
+def filter_translation_words(db_, gl_lang_code, stopwords):
+    '''Filter the translation words from generated stopwords using tws of gl'''
+    tw_lookup = {'hi': 'hi_TTT_1_dictionary', 'ml': 'ml_XYZ_1_dictionary', 'te':
+        'te_XYZ_1_dictionary', 'kan': 'kan_XYZ_1_dictionary', 'ta': 'ta_XYZ_1_dictionary',
+        'mr': 'mr_XYZ_1_dictionary', 'pa': 'pa_XYZ_1_dictionary', 'as': 'as_XYZ_1_dictionary',
+        'gu': 'gu_XYZ_1_dictionary', 'ur': 'ur_XYZ_1_dictionary', 'or': 'or_XYZ_1_dictionary',
+        'bn': 'bn_XYZ_1_dictionary'}
```
kavitharaju, line 175: Even though hard coding this was done as a temporary measure, I would suggest using a version name like "TW" which is more self explanatory and would not collide with test names we might use.

app/crud/nlp_sw_crud.py
```diff
@@ -113,3 +121,208 @@ def add_stopwords(db_: Session, language_code, stopwords_list, user_id=None):
...
+    query = db_.query(db_models.Language.languageId)
+    gl_id = query.filter(func.lower(db_models.Language.code) == gl_lang_code.lower()).first()
+    if not gl_id:
+        raise NotAvailableException("Gateway Language with code %s, not in database. "
+                %gl_lang_code)
+    gl_id = gl_id[0]
+    source_name = tw_lookup[gl_lang_code]
+    response = content_apis.get_dictionary_word(
```
kavitharaju, line 183: Do we need to await this?

app/crud/nlp_sw_crud.py
```diff
@@ -113,3 +121,208 @@ def add_stopwords(db_: Session, language_code, stopwords_list, user_id=None):
...
+        skip=0, limit=100000,
+        db_=db_)
+    if response:
+        translation_words = [row.word for row in response]
+        translation_words = [item.strip() for item in translation_words if item.strip()!='']
+        stopwords = [item for item in stopwords if item[0] not in translation_words]
+    return stopwords
+
+def add_generated_sws(db_, language_id, stopwords, user_id):
+    '''Inserting automatically generated stopwords into db'''
```
kavitharaju, line 197: Can we have a single function for both this and `add_stopwords` defined above?  I feel there might be shared logic here.
This is just a suggestion. Please see if that would be a useful change to make, as less code would mean easier maintenance.

app/crud/nlp_sw_crud.py
anjalyv: We are removing all punctuations and using string.punctuation for that. But string.punctuation has double quotes like " and in our data its different like “ ”. So first i am substituting our quotes with normal quotes and then removing using string.punctuations.

app/crud/nlp_sw_crud.py
```diff
@@ -113,3 +121,208 @@ def add_stopwords(db_: Session, language_code, stopwords_list, user_id=None):
...
+    db_.commit()
+
+def extract_stopwords(db_, language_id, language_code, gl_lang_code, user_id, *args):
+    '''Extract stopwords from available data and stores in db'''
+    sentences = args[0]
+    msg = ''
+    result = []
+    stopwords = find_knee(sentences)
+    if gl_lang_code is None:
+        msg = "Stopwords identified out of limited resources. Manual verification recommended"
```
kavitharaju, line 227: This is a good disclalmer to state. I wonder if we say this based on  sentence count used too

app/crud/nlp_sw_crud.py
```diff
@@ -113,3 +121,208 @@ def add_stopwords(db_: Session, language_code, stopwords_list, user_id=None):
...
+    update_args = {
+                    "status" : schemas_nlp.JobStatus.STARTED.value,
+                    "startTime": datetime.now()
+                   }
+    update_job(db_, job_id, user_id, update_args)
+    language_id = db_.query(db_models.Language.languageId).filter(func.lower(
+                            db_models.Language.code) == language_code.lower()).first()
+    if not language_id:
+        update_args = {
+                    "status" : schemas_nlp.JobStatus.ERROR.value,
```
kavitharaju, line 279: I see we are setting error status for this case explicitly here. What would happen if the operation encounters another error(say DatabaseException for table not found) would that be captured in the jobs table too?

app/db_models.py
```diff
@@ -408,3 +407,14 @@ class StopWords(Base): # pylint: disable=too-few-public-methods
     createdUser = Column('created_user', Integer)
     updatedUser = Column('last_updated_user', Integer)
     updateTime = Column('last_updated_at', DateTime, onupdate=func.now())
+
+class Jobs(Base): # pylint: disable=too-few-public-methods
+    '''Corresponds to table jobs in vachan DB '''
+    __tablename__ = 'jobs'
+
+    jobId = Column('job_id', Integer, primary_key=True, autoincrement=True)
+    userId = Column('user_id', Integer)
```
kavitharaju, line 416: `userId` and other user fields like `createdUser` etc would need String type, I believe, as per the auth module implementations.

app/routers/translation_apis.py
```diff
@@ -417,3 +418,49 @@ def add_stopwords(language_code:schemas.LangCodePattern=Path(...,example="hi"),
     log.debug('language_code:%s, stopwords_list:%s',language_code, stopwords_list)
     msg, result = nlp_sw_crud.add_stopwords(db_, language_code, stopwords_list, user_id=10101)
     return {"message": msg, "data": result}
+
+@router.post('/v2/translation/stopwords/generate',
```
kavitharaju, line 422: This should be `/v2/lookup/stopwords/generate`

app/routers/translation_apis.py
```diff
@@ -417,3 +418,49 @@ def add_stopwords(language_code:schemas.LangCodePattern=Path(...,example="hi"),
...
+    response_model=schemas_nlp.StopWordsGenerateResponse, response_model_exclude_none=True,
+    status_code=200, tags=['Generic Translation'])
```
kavitharaju, line 424: This and other SW apis should be under "lookups" tag

app/routers/translation_apis.py
```diff
@@ -417,3 +418,49 @@ def add_stopwords(language_code:schemas.LangCodePattern=Path(...,example="hi"),
...
+def generate_stopwords(request: Request, background_tasks: BackgroundTasks,
+    language_code:schemas.LangCodePattern=Query(...,example="bi"),
+    use_server_data:bool=True, gl_lang_code:schemas.LangCodePattern=Query(None,example="hi"),
+    sentence_list:List[schemas_nlp.SentenceInput]=Body(None), db_:Session=Depends(get_db)):
+    '''Auto generate stop words for a given language'''
+    log.info('In generate_stopwords')
+    print('In generate_stopwords')
```
kavitharaju, line 431: avoid prints

app/schemas_nlp.py
```diff
@@ -279,3 +279,34 @@ class StopWordsAddResponse(BaseModel):
     '''Response object after adding new stopwords in db'''
     message: str = Field(..., example="3 stopwords added successfully")
     data:List[StopWords] = None
+
+class StopWordsGenerateResponse(BaseModel):
+    '''Response object of auto-generate stopword API'''
+    message: str = Field(..., example="Generating stop words in background")
+    data:dict = Field(..., example={
```
kavitharaju, line 286: Better specify a Job schema class instead of `dict`

app/schemas_nlp.py
```diff
@@ -279,3 +279,34 @@ class StopWordsAddResponse(BaseModel):
...
+        "jobId":100000,
+        "status": "created"})
+
+class JobStatus(Enum):
+    '''Types of job status'''
+    CREATED = 'created'
+    STARTED = 'started'
+    IN_PROGRESS = 'in progress'
```
kavitharaju, line 294: I wonder, would being more explicit as "job created" , "job started" etc be helpful to avoid user thinking SWs created(job completed) when he gets status as just "created"

app/schemas_nlp.py
```diff
@@ -279,3 +279,34 @@ class StopWordsAddResponse(BaseModel):
...
+    FINISHED = 'finished'
+
+class JobCreateResponse(BaseModel):
+    '''Response object of job create api'''
+    message: str = Field(..., example="Job created successfully")
+    data: dict = Field(..., example={
+        "jobId":100000,
+        "status": "created"})
+
+class JobSWOut(BaseModel):
```
kavitharaju, line 306: Avoid SW in the name as this might be used for other kind of background tasks(eg: tokenization) too

app/schemas_nlp.py
```diff
@@ -279,3 +279,34 @@ class StopWordsAddResponse(BaseModel):
...
+    '''Response object for job'''
+    jobId: int = Field(..., example=100000)
+    status: str = Field(..., example="finished")
+    message: str = Field(..., example="Automatically generated stopwords for the given language")
+    languageCode : LangCodePattern =Field(None,example='hi')
+    data:List[StopWords] = None
```
kavitharaju, line 312: Have an output field of dict type and avoid task specific fields languageCode, and data, so that we can have consistent output format for different types of jobs

app/crud/nlp_sw_crud.py
kavitharaju: Using the custom list in utils will do that for you. It has all puncts in `string.punctuation` plus these quotes chars often seen in our data. Maintaining them in one place in code-base would be better than redefining similar things in different places

anjalyv: ok

app/test/test_stop_words_generation.py
anjalyv: assert_positive_response() checks whether response has fields `message` and `data`.  So  assert_positive_response(response.json()) is correct. yeah, status_code should be 201

app/crud/nlp_sw_crud.py
anjalyv: This message is showing if the gl lang is not provided. Means the out is without filtering tws and may contain wrong entries.

kavitharaju:
Merging the PR with the following known issues or unresolved discussions
1. Can we use same function to add SW to DB in add API and generate API? - Not Done
2. Give warning about low resource used based on number of sentences as well - Not Done
3. Would all errors raised during the execution of a task be stored in jobs output - Not answered
4. Change path of SW generate API- Not Done


5. In-order to fix the failing tests in SW, auth has been tweaked in several other places which has now resulted in the failure of multiple tests in sources, dictionaries and get_sentence. This is caused as part of the intergration of auth and SW mudules. These changes will be reverted soon.

Issue #303, opened 2021-12-03
Move access rules to external JSON files


kavitharaju:
See if we can keep the access rules and API_permission_map as two JSON files within the auth module.
Thus we can avoid the use for `access_rules` dictionary and switcher functions in Api_permission_map.py.
Keeping it separate from code would make it easily customizable.

Issue #301, opened 2021-12-03
Keep auth logic as a sub module


kavitharaju:
Move all auth related files to a new directory within app. `vachan-api/app/auth/...`
- authentication.py
- api_permission_map.py(or the JOSN files we are going to add in its place)

schema_auth.py can either be added to this directory or we can create a new directory call schema and all schema files in there, which ever is more easy to import and and manage. I have an inclination to keeping all schema files together in one driectory

Issue #300, opened 2021-12-03
Use logging instead of print statements


kavitharaju:
Change the print statments used in Auth related modules and test to logs.
Also add more logs in other places as required.

Issue #299, opened 2021-12-03
Order of return objects should not be assumed in tests


kavitharaju:
https://github.com/Bridgeconn/vachan-api/blob/0c89853e48c79d68c56e3084c8767dde1bf61b4c/app/test/test_sources.py#L858

As we are not specifying any sort oder in API call, we cannot assume that the response objects will be in a given order. Here in tests, make checks like this instead
```
data = response.json()
for item in data:
   if item['sourceName'] == "ml_TTT_1_bible":
      assert "open-access"] in item['accessPermissions"]
   elif item['sourceName'] == "hi_TTT_1_commentary":
     assert "content" in item["accessPermissions"]
   ....
```

PR #305, opened 2021-12-21
V2 role based access

sijumoncy:
#262 #299 #300 #301 #303 
This PR contains auth and access rules for remaining apis excluding SW for both REST and GraphQL. Fixes of Bugs in test cases.
Auth refactoring and Logging.


app/auth/authentication.py
```diff
@@ -421,13 +371,15 @@ async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-s
         required_params = verify_auth_decorator_params(kwargs)
         db_ = required_params["db_"]
         if required_params['request_context']['endpoint'].startswith("/v2/user"):#pylint: disable=E1126.too-many-nested-blocks
+            log.info('In Auth Decorator - User Specific')
             verified , filtered_content = \
                 check_access_rights(db_, required_params, db_resource)
             if not verified:
                 raise PermissionException("Access Permission Denied for the URL")
```
kavitharaju, line 378: Would be good to have log.error with more details of request for error cases like this

sijumoncy: this log.info is to know the flow of authentication. and in case of error the rasie permsision exception will make a log ?

kavitharaju: I am taking about adding a log.error within the if before we raise the error.
Yes the error handler will make its own logs but will only contain basic details like error message in it, right. This log is to log more details specific to this possibly malicious access. If you think the log within the error handler has enough info in there then its okay

kavitharaju (approved):

Issue #335, opened 2022-02-01
Alter biblevideo tables


kavitharaju:
Alter the DB table schema and apis for bible videos to accommodate the following
* Add a refId_list field to the Bible Videos table, instead of books
* Change the POST, PUT APIs to accept required values and update this field with tagged book, chapter, or verse details of the corresponding video
* Update the GET API to be able to search the tables with 
   * text: doing a fuzzy search in the title, theme, and description fields(like we have in languages table)
   * verse reference : retrieving all videos tagged to the verse, included ranges, chapter or book

PR #341, opened 2022-02-11
Version 2

sijumoncy:
The Bible Project Changes #335


app/routers/content_apis.py
```diff
@@ -702,20 +702,25 @@ async def edit_infographics(request: Request,
 @get_auth_access_check_decorator
 async def get_biblevideo(request: Request,
     source_name:schemas.TableNamePattern=Path(...,example="en_TBP_1_biblevideo"),
-    book_code: schemas.BookCodePattern=Query(None, example="sng"),
     title: str=Query(None, example="Overview: song of songs"),
-    theme: str=Query(None, example="Old Testament"), active: bool=True,
+    series: str=Query(None, example="Old Testament"),
+    search_word: str = Query(None, example="Sri Lanka"),
```
kavitharaju: We could use a different example for search word that makes more sense in the context of Bible Videos.

app/schema/schemas.py
```diff
@@ -1,5 +1,5 @@
 '''Defines all input and output classes for API endpoints'''
-
+#pylint: disable=too-many-lines
```
kavitharaju, line 2: If it is the matter of just one or two lines, we can remove some empty lines or bring up a bracket to adjust that.
If it is more than that consider adding a new schema file. Need not disable it

app/schema/schemas.py
```diff
@@ -917,13 +917,48 @@ class InfographicUpdateResponse(BaseModel):
...
+        '''chapter fields should be greater than or equal to -1'''
+        if val < -1:
+            raise ValueError('chapter field should be greater than or equal to -1')
+        return val
+
 class BibleVideo(BaseModel):
     '''Response object of Bible Vedios'''
     title: str
-    books: List[BookCodePattern]
+    references: List[dict]
```
kavitharaju: Can't we specify a proper class here, instead of dict?

app/crud/contents_crud.py
```diff
@@ -333,8 +333,30 @@ def update_infographics(db_: Session, source_name, infographics, user_id=None):
...
 
-def get_bible_videos(db_:Session, source_name, book_code=None, title=None, theme=None,**kwargs):
+def ref_to_bcv(book,chapter,verse):
+    '''convert reference to BCV format'''
+    bbb = str(book).zfill(3)
+    ccc = str(chapter).zfill(3)
+    vvv = str(verse).zfill(3)
+    return bbb + ccc + vvv
+
+def bcv_to_ref(bcvref,db_):
```
kavitharaju, line 343: Please check if there is similar functonality implemented elsewhere in the code base and make sure we are not redefining existing logic for ref_to_bcv and bcv_to_ref

app/crud/contents_crud.py
```diff
@@ -344,23 +366,60 @@ def get_bible_videos(db_:Session, source_name, book_code=None, title=None, theme
...
+        if book_code and chapter and verse:
+            bcv = ref_to_bcv(book.bookId,chapter,verse)
+            query = query.filter(model_cls.refIds.any(int(bcv)))
+        elif book_code and chapter:
+            code = str(book.bookId)
+            chapter = str(chapter).zfill(3)
+            raw_sql = f'''SELECT * FROM {model_cls.__tablename__}
+                WHERE EXISTS (SELECT 1 FROM unnest(
+                {model_cls.__tablename__}.ref_ids) AS ele WHERE
+                CAST(ele AS TEXT) LIKE '{code+chapter}%')'''
```
kavitharaju: Do we need to cast it to text? Wouldn't performing integer comparisons be more efficient?
Somethig like bbccc000 <= ele < (bbccc000+1000)

app/crud/contents_crud.py
```diff
@@ -344,23 +366,60 @@ def get_bible_videos(db_:Session, source_name, book_code=None, title=None, theme
...
+            id_list = [row[0] for row in result]
+            query = query.filter(model_cls.bibleVideoId.in_(id_list))
+        elif book_code:
+            code = str(book.bookId)
+            code = code+"0" if len(code)==1 else code
+            # query = query.filter(cast(model_cls.refIds,sqlalchemy.String).contains(code+'%'))
+            raw_sql = f'''SELECT * FROM {model_cls.__tablename__}
+             WHERE EXISTS (SELECT 1 FROM unnest(
+                {model_cls.__tablename__}.ref_ids) AS ele
+                WHERE CAST(ele AS TEXT) LIKE '{code}%')'''
```
kavitharaju: Here also, consider performing integer operations instead of casting to text.
bb000000 <= ele <(bb000000+1000000)

app/crud/contents_crud.py
```diff
@@ -344,23 +366,60 @@ def get_bible_videos(db_:Session, source_name, book_code=None, title=None, theme
...
+        search_pattern += ":*"
+        query = query.filter(text("to_tsvector('simple', title || ' ' ||"+\
+            " series || ' ' || description || ' ')"+\
+            " @@ to_tsquery('simple', :pattern)").bindparams(pattern=search_pattern))
+    if book_code:
+        book = db_.query(db_models.BibleBook).filter(
+                db_models.BibleBook.bookCode == book_code.lower() ).first()
+        if book_code and chapter and verse:
+            bcv = ref_to_bcv(book.bookId,chapter,verse)
+            query = query.filter(model_cls.refIds.any(int(bcv)))
```
kavitharaju: This is a usecase to consider...
When we query with a specific verse, do you think we need to return video tagged as related to that whole chapter and whole book too?

app/crud/contents_crud.py
```diff
@@ -371,33 +430,48 @@ def upload_bible_videos(db_: Session, source_name, videos, user_id=None):
...
             if not book:
-                raise NotAvailableException('Bible Book code, %s, not found in database'%book_code)
+                raise NotAvailableException\
+                    ('Bible Book code, %s, not found in database'%buk.bookCode)
+            #generate refid value in BCV
+            if buk.verseEnd is None:
+                buk.verseStart = 0 if buk.verseStart is None else buk.verseStart
+                bcvcode = ref_to_bcv(book.bookId,buk.chapter,buk.verseStart)
+                ref_id_list.add(int(bcvcode))
+            elif not buk.verseEnd is None:
```
kavitharaju: do we need "elif" here? wouldn't "else" be enough?

app/crud/contents_crud.py
```diff
@@ -415,17 +489,29 @@ def update_bible_videos(db_: Session, source_name, videos, user_id=None):
...
-                    raise NotAvailableException('Bible Book code, %s, not found in database'
-                        %book_code )
-            row.books = item.books
-        if item.theme:
-            row.theme = utils.normalize_unicode(item.theme.strip())
+                    raise NotAvailableException\
+                        ('Bible Book code, %s, not found in database'%buk.bookCode)
+                #generate refid value in BCV
+                if buk.verseEnd is None:
+                    buk.verseStart = 0 if buk.verseStart is None else buk.verseStart
```
kavitharaju: This logic seems to be replicated in upload and update. And pylint does make complaints about too many locals and branches. may be we can move this logic out as a function and call it from both places. Consider if that is possible

kavitharaju (changes requested):

app/schema/schemas.py
sijumoncy: it have 50 to 55 line more , if my thought correct

kavitharaju: Then see if you can split the file logically

app/crud/contents_crud.py
```diff
@@ -381,25 +381,28 @@ def get_bible_videos(db_:Session, source_name, book_code=None, title=None, serie
                 db_models.BibleBook.bookCode == book_code.lower() ).first()
         if book_code and chapter and verse:
             bcv = ref_to_bcv(book.bookId,chapter,verse)
-            query = query.filter(model_cls.refIds.any(int(bcv)))
+            fullbook = int(str(book.bookId)+'000000')
+            chapter = str(chapter).zfill(3)
+            book_chapter = int(str(book.bookId)+str(chapter)+"000")
```
kavitharaju: The more efficient way to get the book chapter number from book number and chapter number is this
book_chapter = book.bookId * 1000,000 + chapter*1000

app/crud/contents_crud.py
```diff
@@ -381,25 +381,28 @@ def get_bible_videos(db_:Session, source_name, book_code=None, title=None, serie
...
         elif book_code:
-            code = str(book.bookId)
-            code = code+"0" if len(code)==1 else code
-            # query = query.filter(cast(model_cls.refIds,sqlalchemy.String).contains(code+'%'))
+            code = int(str(book.bookId) + "000000")
             raw_sql = f'''SELECT * FROM {model_cls.__tablename__}
              WHERE EXISTS (SELECT 1 FROM unnest(
                 {model_cls.__tablename__}.ref_ids) AS ele
-                WHERE CAST(ele AS TEXT) LIKE '{code}%')'''
+                WHERE ele >= {code} AND ele < {code+1000000})'''
```
kavitharaju, line 403: I see you have used <= > here and BETWEEN above. Which is better? Use the better one

kavitharaju (changes requested):

kavitharaju (approved):

Issue #347, opened 2022-02-21
Document access management process


kavitharaju:
We are keeping the access tags that are to be attached to resources and their rights, configurable, intended to be maintained by the vachan admin. 
Prepare an instruction document(and a video presentation) that shows how this should be worked on.

Issue #311, opened 2021-12-30
Should "job create" be an API?


kavitharaju:
Related to background tasks(jobs) we have now added two APIs, one for creation and one for status check. The status check would be used by external clients to check on the jobs, but the job creation is, as of now, to be initiated internally by some other APIs triggered by the client(eg: SW generation). So, do we need to expose that job creation function as an API?

kavitharaju:
@joelthe1 , shall we remove this  [job create API](http://api.vachanengine.org/docs#/Jobs/create_job_v2_jobs_post) ?

Issue #306, opened 2021-12-21
Translation words tables are hardcoded in SW operation


kavitharaju:
For the SW identification, we have added provision to use translation words(key terms) to filter out possible content words from the kneed result. For doing this we need to access the TW sources that are available in our DB. They will be added as dictionary tables, whose names will be formed using the version name and revision given for them by the (vachan-)admin who adds data to DB. So we don't have provision to identify them from the other kinds to dictionaries that might be there in the DB. 

For now, we have assumed that the tables will have a predefined name(version and revision) and has hard coded that into code.

kavitharaju:
@joelthe1 , or should we use unfolding word APIs for accessing translationWords ? This is Inidan language(GL) translationWords that are mainly being discussed here.

kavitharaju:
In the SW generate API, instead of specifying a GL, alter it to select a dictionary table from the available sources

PR #355, opened 2022-03-04
Version 2

sijumoncy:
#347  Instruction Document for Access Management
#306  TW GL changed to Source
#311 Job Create Api changed to function


app/crud/nlp_sw_crud.py
```diff
@@ -285,6 +286,27 @@ async def generate_stopwords(db_: Session, request: Request, *args, user_details
     update_job(db_, job_id, user_id, update_args)
     language_id = db_.query(db_models.Language.languageId).filter(func.lower(
                             db_models.Language.code) == language_code.lower()).first()
+
+    if source_name:
+        update_args = {
+                    "status" : schemas_nlp.JobStatus.ERROR.value,
```
kavitharaju, line 292: I don't understand the logic here. Why store as error if `source_name` has a value?

sijumoncy: Source_name is an optional params and the checks for source is only needed when there is source ?

PR #359, opened 2022-03-08
Bugfix/secure kratos admin url

### LINKED ISSUE #352: Block outside access on Kratos admin URL
The Kratos admin URL should be accessible to vachan-api, but not to outside users

kavitharaju:
Attempt to fix #352


kavitharaju:
Merging this now, to test if it works as expected on deployment

Issue #358, opened 2022-03-05
Add auth and access check for GET job status API


kavitharaju:
For now we can set that only a registered and logged in user should be allowed to check the status of a job. Need to define Entitlement(background-task or something) and include resource mapping in code if required, list the API and related operation(read) in the access-permission look up and also define rules in the access-rules file for this.

If required, this could be changed to createdUser. That is, if there is a scenario that the job entry will get deleted or updated in any way when status is checked, then it is better to permit only the createdUser(of that task) to access it.

Issue #336, opened 2022-02-01
Include info of split and merged verse merge in metadata


kavitharaju:
Add a metadata field to the bible cleaned table. Details that could be included in that are
* **Split verses**: we usually combine the split verses and add them as a single verse row. In case of such entries add
```
{"publishedVersification":
    [ 
         {"verseNumber": "3a", "verseText":"...."},
         {"verseNumber": "3b", "verseText":"..."}
    ]}
```
* **Merged verses**: we only keep the verseetext in the first row of the range and keep the remaining rows empty. In such case each row's metadata can hold the same value as below
```
{"publishedVersification":
    [
     {"verseNumber":"4-5", "verseText":"...."}
    ]}
```
* **Versificationn mapping**: In case later we have provision to inlcude versification mapping of the bible to original versification, the info of mapped verses could be captured here
```
{"versificationMap": { "verseNumber": "", "chapter":"", "book":"", "scheme":"org"}}
```
* Additional information in in the USFM that could be useful like **foot notes**, **cross references**

PR #362, opened 2022-03-11
Version 2

sijumoncy:
#336 Split merge verse metadata
#358  get status api auth
TBP Reference Field changes : for [ ]  when the video is about NT, OT etc.


app/crud/contents_crud.py
```diff
@@ -544,35 +545,44 @@ def bible_verse_type_check(content, model_cls_2, book, db_content2, chapter_numb
         match_obj = split_verse_pattern.match(content['verseNumber'])
         post_script = match_obj.group(2)
         verse_number = match_obj.group(1)
+        metadata_field['publishedVersification'].append(
+            {"verseNumber": content["verseNumber"], "verseText":content["verseText"]})
         if post_script == 'a':
```
kavitharaju: Can we avoid this exact matching logic here? What if it is not exactly "a" that is used to denote the first chunk? May be "A" or "i" or even another language letter? 
My suggestion is to use a sorting logic and take the first element of the sorted verseNumbers... Please check if that kind of a logic would work here

kavitharaju (approved):

Issue #363, opened 2022-03-12
List of users: while assigning user to project


udkumar:
As admin on project page I want to assign user to project but I am not able to list of user to select.

Can we add APIs for user list?

kavitharaju:
Users also should have GET and PUT APIs in the same pattern as other contents we have in vachan. These APIs can be implemented such that they call the Kratos APIs from within.

The GET api should be able to filter results based on various available info like roles, active status and also do a partial matching on user-name.

kavitharaju:
In the current implemetation, query parameter `page` is used. Need to change that to `skip` like in other APIs.

PR #364, opened 2022-03-15
Feature/user list

sijumoncy:
#363 Basic API for GET Users List

### DIFF
--- a/app/auth/access_rules.json
+++ b/app/auth/access_rules.json
@@ -57,7 +57,7 @@
         "edit-role":["SuperAdmin"],
        "edit-data":["SuperAdmin", "createdUser"],
         "view-profile":["SuperAdmin", "createdUser"],
-        "read":["SuperAdmin", "AgAdmin", "VachanAdmin"],
+        "read":["registeredUser"],
        "login":["noAuthRequired"],
         "logout":["createdUser"],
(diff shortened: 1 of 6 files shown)


app/auth/access_rules.json
```diff
@@ -57,6 +57,7 @@
         "edit-role":["SuperAdmin"],
        "edit-data":["SuperAdmin", "createdUser"],
         "view-profile":["SuperAdmin", "createdUser"],
+        "read":["SuperAdmin", "AgAdmin", "VachanAdmin"],
```
kavitharaju: All registered users will need to read atleast the name of other users. Example usecase, when a AgUser who is a project owner wants to add another user to the project

app/routers/auth_api.py
```diff
@@ -49,6 +51,21 @@ async def login(user_email: str,password: types.SecretStr,
     log.debug('login:%s',user_email)
     return user_login_kratos(user_email,password)
 
+@router.get('/v2/user/list-identities', response_model=List[schema_auth.IdentitityListResponse],
```
kavitharaju: Can we keep the URL simple and similar to our other GET endpoints?
Something like `/v2/users`

kavitharaju (changes requested):

kavitharaju (approved):

Issue #353, opened 2022-03-01
Update Languages Data in DB


kavitharaju:
Ensure the languages data we have currently in the DB is updated as per the sources
- [language-subtag-registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry)
- UnfoldingWord's GL research database

PR #365, opened 2022-03-18
Feature/users

sijumoncy:
1.  #363 Kratos Users Endpoint (get user/users (with filtering params) , put user )
2. #353 DB language Data modified and updated the Language file and script
3.  Schema change for Login response added a new field to return user Id


app/auth/authentication.py
```diff
@@ -231,12 +231,19 @@ async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-s
...
         authenticated = check_right(user_details, required_rights)
+
         if (resource_type == schema_auth.ResourceType.USER and not authenticated):
             # Need to raise error before function execution, as we cannot delay db commit
             # like we do in other cases as changes happen in Kratos db, not app db'''
             if user_details['user_id'] is None:
                 raise UnAuthorizedException("Access token not provided or user not recognized")
-            raise PermissionException("Access Permission Denied for the URL")
+            #check for created user
+            if method == "PUT" and len(path_params)>0 :
```
kavitharaju: Do we need to check this here like this? Wouldn't the mapping to "PUT" method, requiring created user access be implemeneted via the api_permission map and rules.json?

app/auth/authentication.py
```diff
@@ -319,43 +326,110 @@ def kratos_logout(recieve_token):
...
+                kratos_role = [x.lower() for x in data["traits"]["userrole"]]
+                for k_role in kratos_role:
+                    res = list(filter(k_role.startswith, role_list)) != []
+                    if res:
+                        role_status = True
+                        break
+                    role_status = False
+            if name_status and role_status:
+                user_data.append(kratos_user)
+        user_data = user_data[skip:skip+limit] if skip>=0 and limit>=0 else []
```
kavitharaju, line 374: are we fetching all users from Kratos and then implementing pagination with our code? can't we translate the skip togic we have to their page logic and request them for users passing page

kavitharaju (approved):

Issue #356, opened 2022-03-04
Access resources from gitlab repositories


kavitharaju:
Connect securely to the self hosted gitlab instance and access media resources stored there so that API requests for them from clients applications can be served

kavitharaju:
Add endpoints 
* /v2/media/gitlab/stream
* /v2/media/gitlab/download

Need to implement auth mechanism
* Create a content type "gitlab-repo"
* Define operations like "stream", "download" etc and define access rules and API permission maps
* Make necessary changes in auth module to handle the new content-type and resource type

In the streaming API, consider if we can implement the following
* accepting repo+tag+file_path or permanent_link to identify a resource. 
* accepting start_time and end_time, to access portions of audio or video. 
* effective way to cache media fetched from gitlab(look into redis)
* reduce the time delay in first response, due to the time taken to fetch the full content from gitlab

PR #369, opened 2022-04-12
Feature/mediastream

sijumoncy:
#356 
PR contents
- Stream Gitlab repo video
- Download Gitlab repo video
- Authentication based on source permission (source per repo with metadata have repo url)


app/.pylintrc
```diff
@@ -153,7 +153,8 @@ disable=print-statement,
         xreadlines-attribute,
         deprecated-sys-function,
         exception-escape,
-        comprehension-escape
+        comprehension-escape,
+        R0801
```
kavitharaju: I recommend not disabling a rule globally like this. If you feel the need to suppress this warning, do that in a smaller scope. At a particular location in code, in a sub-module or may be in a module if that is a recurring case and which must be suppressed. But not globally like this.

Especially this warning, which identifies duplicate code, is one that would help us with maintaining code quality and avoid redundancy.

app/crud/media_crud.py
```diff
@@ -0,0 +1,102 @@
...
+BYTES_PER_RESPONSE = 100000
+CACHEDMEDIA = []
+cached_media_download = []
+MEDIA_CACHE_LIMIT = 3
+
+gl = gitlab.Gitlab(url="https://gitlab.bridgeconn.com", private_token=access_token)
+
+def media_streamer(stream, chunk_size, start, size):
+    '''chunk up the object and return in pieces'''
+    print("comes in streamer once with start:", start)
```
kavitharaju: Could removeor comment out these print statements

app/db_models.py
```diff
@@ -24,6 +24,7 @@ class ContentTypeName(Enum):
     INFOGRAPHIC = "infographic"
     BIBLEVIDEO = "biblevideo"
     DICTIONARY = "dictionary"
+    GITLABREPO = "gitlab-repo"
```
kavitharaju: I think we could use `_` instead of `-` or not use any separator between the two words. This is just out of concern of aesthetics of the source name. As we use _ to connect other parts of the source name such as language version etc, having just one "-" at the end could be avoided.

My suggestion is not to use any separator between and just call it "gitlabrepo", as there might be places in code where might split the source name with "_" to get the constituents.

app/routers/media_api.py
```diff
@@ -0,0 +1,121 @@
...
+    404:{"model": schemas.ErrorResponse},403:{"model": schemas.ErrorResponse}},
+    status_code=200, tags=["Media"])
+@get_auth_access_check_decorator
+async def stream_media(request: Request, #pylint: disable=unused-argument,too-many-arguments
+    repo: str = Query(None,example="kavitha.raju/trial-media-project"),
+    tag: str = Query(None,example="main"),
+    file_path: str=Query(None,example="token videos/Apostle.MOV"),
+    permanent_link: str=Query(None,example=
+        "https://gitlab.bridgeconn.com/kavitha.raju/\
+            trial-media-project/-/raw/main/token videos/Apostle.MOV"),
```
kavitharaju: A permalink would have a commit hash in place of main or other branch names. This is an example
https://gitlab.bridgeconn.com/kavitha.raju/trial-media-project/-/blob/e79238bbf860c709dc31d31e69d3416587c4614d/Docker_in_VE.mp4

app/routers/media_api.py
```diff
@@ -0,0 +1,121 @@
...
+    * to access a content provide either (repo + tag+ file_path) or permanent_link
+    * permanent link will be considered if its given
+    * start and end time are optional in format of H:M:S.000
+    '''
+    log.info('In get_stream_media')
+    log.debug('repo:%s, tag %s, file_path: %s, permanent_link: %s, start_time: %s, end_time: %s',
+        repo, tag, file_path, permanent_link, start_time, end_time)
+    if not permanent_link:
+        if not repo or not file_path:
+            raise UnprocessableException("Either Permanent Link or repo + file_path")
```
kavitharaju: The message could be "Either Permanent Link or repo + file_path is madatory to identify the media"

app/routers/media_api.py
```diff
@@ -0,0 +1,121 @@
...
+    db_source = media_crud.find_media_source(repo, db_)
+    if db_source is None:
+        raise NotAvailableException(f"No source is available for {repo}")
+    source_name = db_source.sourceName
+    request.scope['path'] = "/v2/sources"
+    request._url = URL("/v2/sources")#pylint: disable=W0212
+
+    tables = await get_source_and_permission_check(source_name, request, user_details, db_)
+    if len(tables) == 0:
+        raise NotAvailableException("No sources available for the requested name or language")
```
kavitharaju: Just add a check for `len(tables) > 1` also. In case same repo is created as multiple sources and different permissions. 

This is to be checked for and  avoided at source creation as well. I dont know if we can make unique constraint for repo-name at DB level, as we are using JSON field. But it needs to be checked in code at least

app/routers/media_api.py
```diff
@@ -0,0 +1,121 @@
...
+    log.debug('repo:%s, tag %s, file_path: %s, permanent_link: %s, start_time: %s, end_time: %s',
+        repo, tag, file_path, permanent_link, start_time, end_time)
+    if not permanent_link:
+        if not repo or not file_path:
+            raise UnprocessableException("Either Permanent Link or repo + file_path")
+        repo = "https://gitlab.bridgeconn.com/" + repo
+    else:
+        repo = permanent_link.split("/-/")[0]
+        part2 = permanent_link.split("/-/")[1]
+        tag =  re.search(r'[^raw/]\w+',part2)[0]
```
kavitharaju: I have doubt about this... Need to make sure if a link will always have `raw` in it

app/routers/media_api.py
```diff
@@ -0,0 +1,121 @@
...
+        file_path = re.search(f'[^raw/{tag}].+',part2)[0]
+
+    # find source
+    db_source = media_crud.find_media_source(repo, db_)
+    if db_source is None:
+        raise NotAvailableException(f"No source is available for {repo}")
+    source_name = db_source.sourceName
+    request.scope['path'] = "/v2/sources"
+    request._url = URL("/v2/sources")#pylint: disable=W0212
```
kavitharaju: I wonder, is this how we have implemented the auth check in other cases where we have same API accessing multiple resources?

I think we were adding another entry in the API-permissions like
```"/v2/media/gitlab/stream", None, False/True, differnent-resource-types, stream```

File: app/routers/media_api.py [hunk 9]
```diff
@@ -0,0 +1,121 @@
...
+
+
+@router.get("/v2/media/gitlab/download",
+    responses={502: {"model": schemas.ErrorResponse},
+    422: {"model": schemas.ErrorResponse},401:{"model": schemas.ErrorResponse},
+    404:{"model": schemas.ErrorResponse},403:{"model": schemas.ErrorResponse}},
+    status_code=200, tags=["Media"])
+@get_auth_access_check_decorator
+async def download_media(request: Request, #pylint: disable=too-many-arguments
+    repo: str = Query(None,example="kavitha.raju/trial-media-project"),
```
kavitharaju, line 103: I see a lot of logic related to identifying repo, chekcing access etc are same for both these APIs. Consider moving the common parts to a different function and avoid redundancy

app/routers/utils.py
```diff
@@ -0,0 +1,26 @@
+# pylint: disable=R0801
+'''Utility functions'''
+from dependencies import log
+from schema import schema_auth
+from routers.content_apis import get_source
+
+async def get_source_and_permission_check(source_name, request, user_details, db_):
+    """get source with acess right check"""
```
kavitharaju: Would it be better to add one more optional parameter `source_name` to the `get_source()` router function than writing a new method to that?

app/schema/schemas.py
```diff
@@ -314,7 +314,7 @@ class Config:
             }
         }
 
-TableNamePattern = constr(regex=r"^[a-zA-Z]+(-[a-zA-Z0-9]+)*_[A-Z]+_\w+_[a-z]+$")
+TableNamePattern = constr(regex=r"^[a-zA-Z]+(-[a-zA-Z0-9]+)*_[A-Z]+_\w+_[a-z,-]+$")
```
kavitharaju: comma is not required.
As per my above comment, if you agree to that, `-` could also be avoided

requirements.txt
```diff
@@ -1,13 +1,20 @@
+aiofile==3.7.4
+aiofiles==0.8.0
```
kavitharaju, line 2: Did you have to update any library for adding these features?

sijumoncy: this need for download api

app/crud/structurals_crud.py
```diff
@@ -198,31 +198,33 @@ def get_sources(db_: Session,#pylint: disable=too-many-locals,too-many-branches,
...
-        query = query.filter(db_models.Source.license.has(code = license_abbreviation.strip()))
-    if language_code:
-        query = query.filter(db_models.Source.language.has(code = language_code.strip()))
-    if metadata:
-        meta = json.loads(metadata)
-        for key in meta:
-            query = query.filter(db_models.Source.metaData.op('->>')(key) == meta[key])
+    if source_name:
+        query = query.filter(db_models.Source.sourceName == source_name)
+    else:
```
kavitharaju: I know that asp er our implementation logic source_name uniquely identifies an item and there is no point in addinging additional filters if there is source_name. But the client need not know this. what is he gives a source_name and another filter,, say a version or metadata or access_tag. Then if we dont use those additional params in the filter the API will return the item matching the source_name and other inputs need not match. This might create an issue. 

So, if there are no other problems I would suggest not using this `else`.Just add one filter for source_name if input is there

kavitharaju (commented):
Other than the one issue I mentioned all looks good

kavitharaju (approved):

Issue #377, opened 2022-05-02
Improvements on media API


kavitharaju:
1. gitlabrepo, as a contenttype, should be supported by default. So include that in the seed DB
2. now the API returns unsupported content-type error for mp3. Only mp4 and mov are supported. Need to include support for all common media types(audio, video, image)
3. Need to revisit the authentication rules configured for media API. While working on it, I could access the API with SA credentials while a regular API-user was not able to. Need to verify if this behavior is as per the configured rules.
4

PR #378, opened 2022-05-03
Feature/mediastream

sijumoncy:
#377 

- Content Type for medias ( Video, Audio, Image)
- Header Range Error when request as normal api call ( Not From Player or stream supported methods)
- Manual Check Done for Stream and Download apis for Video, Audio and Image
- Auth verified as per rule ( Manually )

### DIFF
--- a/app/crud/media_crud.py
+++ b/app/crud/media_crud.py
@@ -1,7 +1,9 @@
 ''' Place to define all data processing and operations
 related to gitlab media operations'''
 import os
+import mimetypes
 from datetime import datetime
+from fastapi import HTTPException
 from fastapi.responses import StreamingResponse, Response
 import gitlab
 import db_models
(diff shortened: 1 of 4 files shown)


app/crud/media_crud.py
```diff
@@ -25,22 +27,26 @@ def get_gitlab_stream(request, repo, tag, file_path,permanent_link,**kwargs):#py
...
-        content_type =  "video/mp4"
-    elif url.endswith("mov") or url.endswith("MOV"):
-        content_type =  "video/quicktime"
-    else:
-        raise Exception("Unsupported video format!")
+    content_type = mimetypes.guess_type(url.split("/")[-1], strict=True)
+    if content_type is None:
+        raise Exception("Unsupported media format!")
+
+    if "image" not in content_type[0]:
```
kavitharaju: Hmm.. now I wonder if image needs to be brought under this API...

app/crud/media_crud.py
```diff
@@ -25,22 +27,26 @@ def get_gitlab_stream(request, repo, tag, file_path,permanent_link,**kwargs):#py
...
     else:
         url = permanent_link
 
-    if url.endswith("mp4"):
-        content_type =  "video/mp4"
-    elif url.endswith("mov") or url.endswith("MOV"):
-        content_type =  "video/quicktime"
-    else:
-        raise Exception("Unsupported video format!")
+    content_type = mimetypes.guess_type(url.split("/")[-1], strict=True)
```
kavitharaju, line 38: This is the better way to do it

kavitharaju (approved):
The implemented changes looks good. 
There was a mention of including gitlabrepo as a contnettype in the seed DB in the issue. Is that done? Will it be reflected upon a fresh DB creation?

What is the auth congiuration now? Can any API-user(or unregistered user) access the API, when the content is marked as open-access?

sijumoncy:
- Removed Image from stream ( Download will support it . Manually tested)
- Auth test done manually ( Api user access open-access content)
- test cases updated for errors 
- mimetypes is used based on most of the suggestions found  to work with media content type in python

kavitharaju (approved):

PR #388, opened 2022-06-27
Feature/media reddis

sijumoncy:
- Container For Redis DB cache

PR #394, opened 2022-07-06
Bugix/redis test failure

kavitharaju:
- Includes Redis container additions from PR #388 
- Resolves the test failure in the above PR by updating the code
   - tries to ping redis
   - if not connected, work without using cache, instead of raising error
   - this comes useful in local development as well as running tests on git actions
- Removes the port mapping of redis container to host port. This would resolve the security issue of open redis port DO alerted about
- Code clean up in media module
    - remove the old caching logic of using a python list variable
    - uses env variables for Redis host, port etc rather than hard coding
    - make Redis connection require a password
    - removed reduntant Response object creation in media download (in crud function and router function)
    - removed unnecessary gitlab download before the streaming logic in streaming router function
 - Changed indentation in docker-compose to allow proper code folding in IDE


kavitharaju:
@samuelanthrayose  the changes in this PR would resolve the security issue DO reported. You can review if you would like to

sijumoncy (approved):

PR #397, opened 2022-07-08
Feature/project books update upon csv upload

### LINKED ISSUE #393: Uploaded file using csv but available book list not getting updated.
Hi,
was adding books via csv , got success message and  was able to see the data in the payload but not able to see any book in response.
eg ProjectID- 100010 had added book MAT via csv , but Mat is not reflecting in the book names. had conirmed the that content is there via sentence api.
attaching screenshot for the same
Screenshot from 2022-07-05 18-52-24

kavitharaju:
fixes #393 
When source sentences are upload as `sentenceList` (instead of USFM or selecting from DB itself), now the server code checks if `surrogateId` of sentences starts with book code, thus identify the books and add the information to the project metadata.

### DIFF
--- a/app/crud/projects_crud.py
+++ b/app/crud/projects_crud.py
@@ -48,7 +48,7 @@ def create_agmt_project(db_:Session, project, user_id=None):
     # db_.commit()
     return db_content
 
-book_pattern_in_surrogate_id = re.compile(r'^\w\w\w')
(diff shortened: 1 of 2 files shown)


LijuJacob08:
Hi Kavitha, 
had tried uploading the csv file(tried it via api as well via app),  book list was  getting updated .

kavitharaju:
Thanks for confirming

Issue #398, opened 2022-07-14
server crashing while making suggestion  API call


LijuJacob08:
Hi Kavitha,

1 facing this issue while opening the Project page which is initiating the suggestion API . 

----------------------------------------------
Observations
- Project opens for the book size is small(like having 1 chapter with few verse) in a minute or two. 
- noted that suggestion api is being called twice

following are the request details:-

1)Request Header

    1. :authority: 
       api.vachanengine.org
    2. :method: 
       PUT
    3. :path: 
       /v2/autographa/project/suggestions?project_id=100016
    4. :scheme: 
       https
    5. accept: 
       */*
    6. accept-encoding: 
       gzip, deflate, br
    7. accept-language: 
       en-US,en;q=0.9
    8. app: 
       Autographa
    9. authorization: 
       Bearer zLLme29wyO8uRr8nHSAblDIoyNCIo2Z3
    10. content-length: 
       0
    11. origin: 
       http://localhost:3000
    12. referer: 
       http://localhost:3000/
    13. sec-ch-ua: 
       " Not A;Brand";v="99", "Chromium";v="100", "Google Chrome";v="100"
    14. sec-ch-ua-mobile: 
       ?0
    15. sec-ch-ua-platform: 
       "Linux"
    16. sec-fetch-dest: 
       empty
    17. sec-fetch-mode: 
       cors
    18. sec-fetch-site: 
       cross-site
    19. user-agent: 
       Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.75 Safari/537.36

2)Payload
![image](https://user-images.githubusercontent.com/93033055/178917640-ca95f2e1-f163-4b47-8927-18f6e55cc86e.png)

3)suggestion was called thrice(3=1 options +2 PUT calls)

![image](https://user-images.githubusercontent.com/93033055/178917728-3d746119-242d-4239-9da4-56a338e2b59e.png)

4)after some time got 

![image](https://user-images.githubusercontent.com/93033055/178917757-2a549b1c-f30f-4757-8e97-eb7ba88ca903.png)
-----------------------------------------------------------------------------------

kavitharaju:
Why is the same API called thrice?

Also, are you not using query params like `sentence_id_list` or `sentence_id_range`? This specifies the verses you are trying to load now. If not specified, the suggestion generation will be performed on all available verses in the project, which will be resource intense.

[...]

PR #401, opened 2022-07-27
Feature/cache tries

### LINKED ISSUE #392: Redis to cache the gloss/token tries
Check possibility of using Redis to cache the gloss/token tries. 

Now a trie is created every time a tokenization request is made. This leads to long response time for the tokenization APIs and other APIs that use tokenization internally. If the trie is once made and loaded to cache, this time cam be reduced.

The tries stored in cache also needs to be updated upon addition of new tokens to the translation memory via gloss API or token translation API

kavitharaju:
fixes #392 , in the context of  #398 
Stores the tries in the cache memory to avoid re building it every time we we call tokenization or suggestion APIs
Updated the trie in cache memory if there are new tokens added to translation memory

### DIFF
--- a/app/crud/nlp_crud.py
+++ b/app/crud/nlp_crud.py
@@ -4,6 +4,7 @@ related to NLP operations and translation apps'''
 import re
 import os
 import json
+import pickle
 from datetime import datetime
 from math import floor, ceil
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


kavitharaju:
The effect of this change needs to be tested
https://github.com/Bridgeconn/vachan-api/issues/392#issuecomment-1196486599

kavitharaju:
Tests show this solution is feasible
https://github.com/Bridgeconn/vachan-api/issues/392#issuecomment-1228311011

PR #410, opened 2022-08-12
Feature/offenbackup

sijumoncy:
- Docker Volume Backups with Offen

### DIFF
--- a/docker/Kratos_config/database.yml
+++ b/docker/Kratos_config/database.yml
@@ -31,6 +31,65 @@ services:
       - POSTGRES_HOST_AUTH_METHOD=md5
     networks:
       - vachan-auth-net
+  
+  # offen backup
+  offen-backup_kratos_daily: &backup_service
+    image: offen/docker-volume-backup:latest
+    restart: always
+    environment: &backup_environment
+      #Daily 00 : 00
+      BACKUP_CRON_EXPRESSION: "0 0 * * *"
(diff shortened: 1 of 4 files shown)


docker/Kratos_config/database.yml
```diff
@@ -31,7 +31,84 @@ services:
...
+    image: offen/docker-volume-backup:latest
+    restart: always
+    environment: &backup_environment #Daily 00 : 00
+      BACKUP_CRON_EXPRESSION: "0 0 * * *"
+      BACKUP_FILENAME: backup-kratos-daily-%Y-%m-%dT%H-%M-%S.tar.gz
+      BACKUP_SOURCES: /docker_kratos-sqlite
+      BACKUP_PRUNING_PREFIX: backup-kratos-daily-
+      BACKUP_RETENTION_DAYS: 5
+    depends_on:
+      - kratos-migrate
```
kavitharaju, line 46: should it be depending on migrate or Krato's DB container?

sijumoncy: migrate is depend on kratos DB (postgres). is that enough ?

kavitharaju: It must be enough. My concern is because the migrate container is one that stops after it is run. Hope it wont have any effect on ensuring these offen containers are running. I read that the `depends_on` just determines the order in whichh it is started, so it must be okay.

PR #450, opened 2022-11-08
Implement Hard Delete on Languages and Content Types

shimilgithub:


.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
sijumoncy, line 1: why we have an xml file here ?

app/auth/access_rules.json
```diff
@@ -5,8 +5,9 @@
         "read-via-api":["noAuthRequired"],
         "read-via-vachanadmin":["SuperAdmin", "VachanAdmin"],
         "view-on-web":["noAuthRequired"],
-        "refer-for-translation": ["registeredUser"]
-    },
+        "refer-for-translation": ["registeredUser"],
+        "delete/deactivate":["registeredUser"]
```
sijumoncy, line 9: Do we allow to delete meta-content by any registered user or resourceCreatedUser , superAdmin as like in edit

app/auth/api-permissions.csv
```diff
@@ -12,12 +12,15 @@ Endpoints,Method,RequestApp,FilterResults,ResourceType,Permission
...
 /v2/contents,POST,None,False,None,create
+/v2/contents,DELETE,None,False,None,delete/deactivate
 /v2/languages,GET,Autographa,False,None,refer-for-translation
 /v2/languages,GET,Vachan-online or vachan-app,False,None,view-on-web
 /v2/languages,GET,VachanAdmin,False,None,read-via-vachanadmin
 /v2/languages,GET,API-user,False,None,read-via-api
 /v2/languages,POST,None,False,None,create
 /v2/languages,PUT,None,False,None,edit
+/v2/languages,DELETE,None,False,None,delete/deactivate
+/v2/restore,POST,None,False,"Manipulate Data",data-restore
```
sijumoncy, line 23: v2/restore is POST or PUT , I think its put because its not a new data creation , restore data is already existed one. clarify with @kavitharaju  too

sijumoncy: "Manipulate Data" in the resource column instead of None is requiered ? , I can not find this value is passing in operates on field and the resource type is identified by splitting the enpoint.

app/crud/structurals_crud.py
```diff
@@ -30,6 +33,14 @@ def create_content_type(db_: Session, content: schemas.ContentTypeCreate):
     # db_.refresh(db_content)
     return db_content
 
+def delete_content(db_: Session, content: schemas.ContentIdentity):
```
sijumoncy, line 36: Do we need separate schemas for each meta content or can we use a common schema with id : int  for all, because the delete is based on the unique Id  and the table is specifying in the respective crud delete.

app/crud/structurals_crud.py
```diff
@@ -89,6 +100,51 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
     # db_.refresh(db_content)
     return db_content
 
+def add_deleted_data(db_: Session, del_content : db_models.Language,
```
sijumoncy, line 103: this function is a common function for adding delete content to deleted table . if so, why used specific db model for del_content

app/crud/structurals_crud.py
```diff
@@ -89,6 +100,51 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
+    #db_.commit()
+    return del_content
+
+def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
+    '''Restore deleted record back to the original table'''
+    restore_content = db_.query(db_models.DeletedItem).get(restored_item.itemId)
+    db_content = restore_content
+    json_string = db_content.deletedData
+    db_.delete(restore_content)
+    db_.commit()
```
sijumoncy, line 123: what will happen if the auth failed . the data will loss from delete table ?

app/crud/structurals_crud.py
```diff
@@ -89,6 +100,51 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
+    if db_content.deletedFrom == 'content_types':
```
sijumoncy, line 124: I feel the if elif ladder will make this function too long in the future , when we implement restore for all db tables.

app/crud/structurals_crud.py
```diff
@@ -89,6 +100,51 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
+        languageId = json_string['languageId'],
+        metaData= json_string['metaData'],
+        createdUser= json_string['createdUser'],
+        updatedUser= json_string['updatedUser'],
+        updateTime = datetime.now())
+    db_.add(db_content)
+    #db_.commit()
+    return db_content
+
+def delete_language(db_: Session, lang: schemas.LanguageIdentity):
```
sijumoncy, line 140: is it possible to have a common function for delete content from db, if  we pass current db model and id as params to the functions

app/deleted_contents.json
```diff
@@ -0,0 +1 @@
+"{\"contentType\": \"contente\", \"contentId\": 100028}"
```
sijumoncy, line 1: why we need a JSON , do we have a requirement to keep deleted items data in json. I am not able to find how this json is created ?

app/routers/content_apis.py
```diff
@@ -57,6 +58,43 @@ async def add_contents(request: Request, content: schemas.ContentTypeCreate,
     return {'message': "Content type created successfully",
             "data": data}
 
+@router.delete('/v2/contents',response_model=schemas.ContentDeleteResponse,
+    responses={404: {"model": schemas.ErrorResponse},
+    401: {"model": schemas.ErrorResponse},422: {"model": schemas.ErrorResponse}, \
+    502: {"model": schemas.ErrorResponse},409: {"model": schemas.ErrorResponse}},
```
sijumoncy, line 64: 409 - conflict - is there a possibility of having conflict in delete action ?

app/routers/content_apis.py
```diff
@@ -57,6 +58,43 @@ async def add_contents(request: Request, content: schemas.ContentTypeCreate,
...
+    status_code=200,tags=["Contents Types"])
+@get_auth_access_check_decorator
+async def delete_contents(request: Request, content_obj: schemas.ContentIdentity = Body(...),
```
sijumoncy, line 67: pass id as body param or query param, which is better ?

app/routers/content_apis.py
```diff
@@ -57,6 +58,43 @@ async def add_contents(request: Request, content: schemas.ContentTypeCreate,
...
+    delcont = structurals_crud.add_deleted_data(db_=db_,del_content= deleted_content,
+             user_id=user_details['user_id'],table_name = dbtable_name)
+    return {'message': f"Content with identity {content_id} deleted successfully",
+            "data": delcont}
+
+@router.post('/v2/restore', response_model=schemas.DataRestoreResponse,
+    responses={502: {"model": schemas.ErrorResponse}, \
+    422: {"model": schemas.ErrorResponse},401:{"model": schemas.ErrorResponse},
+    409: {"model": schemas.ErrorResponse}},
+    status_code=201, tags=["Contents Types"])
```
sijumoncy, line 85: restore is a common functionality endpoint. is it better if we add this as separate in swagger other than under the content - type

app/schema/schemas.py
```diff
@@ -129,6 +130,51 @@ class Config:
             }
         }
 
+class LanguageDeleteResponse(BaseModel):
+    """Language delete response"""
+    message:str
+
+class LanguageIdentity(BaseModel):
+    """kratos language ID input"""
```
sijumoncy, line 138: why it is kratos ?

app/schema/schemas.py
```diff
@@ -129,6 +130,51 @@ class Config:
...
+    class Config:
+        '''display example value in API documentation'''
+        schema_extra = {
+            "example": {
+                "languageId": 100057
+            }
+        }
+
+class RestoreIdentity(BaseModel):
+    """kratos item ID input"""
```
sijumoncy, line 149: kratos is mentioned in couple of schemas. ?

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
     response = client.post(UNIT_URL, headers=headers, json=data)
     assert_input_validation_error(response)
 
+def populate_db(db_ : Session): # pylint: disable=C0116
```
sijumoncy, line 100: you can call create content type api to add test content as in the other tests

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
...
+    VALUES ('testdata')''')
+    db_.commit()
+    content_id = db_.execute('''SELECT content_type_id FROM content_types \
+    WHERE (content_type = 'testdata')''')
+    db_.commit()
+    [test_content_id] = content_id.fetchone()
+    test_id = test_content_id
+    return test_id
+
+def delete_db(db_ : Session): # pylint: disable=C0116
```
sijumoncy, line 111: call delete api

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
...
+
+def test_delete_notavailable_content_id():
+    ''' request a non existing content ID, Ensure there is no partial matching'''
+    data = {"contentId":1000}
+    headers = {"contentType": "application/json",
+                "accept": "application/json",
+                'Authorization': "Bearer"+" "+initial_test_users['APIUser2']['token']
+            }
+    response = client.delete(UNIT_URL,headers=headers,json=data)
+    assert response.status_code == 502
```
sijumoncy, line 197: not available should return 404 (Not Found ) . right ?

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
...
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+token_admin
+                     }
+    response = client.post(RESTORE_URL, headers=headers_admin, json=data)
+    assert response.status_code == 201
+    assert response.json()['message'] == \
+    f"Deleted Item with identity {item_id} restored successfully"
+    logout_user(token_admin)
+
+def test_restore_incorrectdatatype():
```
sijumoncy, line 258: you can add tests for restore contents with other users , Aguser, VachanUser etc.

app/test/test_languages.py
```diff
@@ -341,8 +340,111 @@ def test_searching():
...
         if lang['language'] == "Sinhala":
             found = True
-    assert found    
+    assert found
 
     # ensure search words are not stemmed as per rules of english
     response = client.get(UNIT_URL+"?search_word=chinese")
     assert len(response.json()) > 5
+
+def populate_language_db(db_ : Session): # pylint: disable=C0116
```
sijumoncy, line 349: you can add restore test for languages

db/seed_DB.sql
```diff
@@ -38,7 +38,7 @@ CREATE INDEX languages_search_idx ON languages USING gin (
 );
 
 
-\COPY languages (language_code,language_name, script_direction, metadata) FROM 'csvs/consolidated_languages.csv' DELIMITER ',' CSV;
+\copy languages (language_code,language_name, script_direction, metadata) FROM 'csvs/consolidated_languages.csv' DELIMITER ',' CSV;
```
sijumoncy, line 41: why edited seed DB ?

docker/vachan-ENV/pyvenv.cfg
```diff
@@ -0,0 +1,3 @@
+home = /usr/bin
```
sijumoncy, line 1: the above env files are not required

package-lock.json
```diff
@@ -0,0 +1,10 @@
+{
```
sijumoncy, line 1: why we have package json lock ?

app/crud/structurals_crud.py
shimilgithub: Data won't loss.Person authorised as Super Admin can only perform restore.

app/crud/structurals_crud.py
shimilgithub: Yes.Since there is a single API for restoring to any table,conditional statement will be lengthy. Else I need to keep separate restore API for each table.

app/deleted_contents.json
shimilgithub: This file added by mistake.WIll remove it

app/routers/content_apis.py
shimilgithub: OK.What would be the tag for restore?

app/crud/structurals_crud.py
sijumoncy: There are 2 auth flow , 
1. authenticate and call crud function
2. call crud function and authenticate 
if the data will not loss means this in first category. right ?

app/crud/structurals_crud.py
sijumoncy: Is it possible to pass the db and other required params to a common function

app/auth/access_rules.json
shimilgithub: shall I add resourceCreatedUser  and superAdmin also?

.project
kavitharaju: yeah need to remove this

app/auth/access_rules.json
kavitharaju: should be createdUser and SuperAdmin only. So we can prevent users deleting contents preloaded in seed DB.

app/auth/access_rules.json
```diff
@@ -75,5 +76,8 @@
        "login":["noAuthRequired"],
         "logout":["createdUser"],
         "delete/deactivate":["SuperAdmin"]
```
kavitharaju, line 78: Here also it should be createdUser and SuperAdmin

app/crud/structurals_crud.py
shimilgithub: yes.

app/auth/api-permissions.csv
kavitharaju: I am not very particular on POST or PUT, but I think PUT will be better.

kavitharaju: Keeping data-manipulate in CSV may not be required, but I guess no harm in doing that. But the operates on might be required on the router function of restore.

app/crud/structurals_crud.py
kavitharaju: I agree to Siju's comment. It will save us a lot of code lines.

app/crud/structurals_crud.py
kavitharaju: Agree.

app/crud/structurals_crud.py
kavitharaju: yeah. db_.commit() should not be here I think

app/auth/access_rules.json
sijumoncy: confirm with kavitha, I think one created the content and super admin can delete the content, not every user

app/crud/structurals_crud.py
kavitharaju: Shimil can you write a single SQLAlchemy statement(or SQL statement as last resort), that takes the table to insert as a variable. This variable can be the table name you get from the deletedtable's record. And columns can somehow be understood from the contents of the JSON field. Try if it is possible, because it would solve all types of table in one go.

app/crud/structurals_crud.py
sijumoncy: addd test cases for restore with users - api, vachan, vachanadmin, aguser, agadmin etc. will help to avoid confusions in future

app/crud/structurals_crud.py
```diff
@@ -89,6 +100,51 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
+        deletedTime = datetime.now(),
+        deletedFrom = table_name)
+    db_.add(db_content)
+    #db_.commit()
+    return del_content
+
+def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
+    '''Restore deleted record back to the original table'''
+    restore_content = db_.query(db_models.DeletedItem).get(restored_item.itemId)
+    db_content = restore_content
```
kavitharaju, line 120: Why do we need two variables `db_content` and `restore_content`? Wouldn't one suffice?

app/crud/structurals_crud.py
kavitharaju: This is not a must, as we have separate crud functions for insert and edit. But unlike insert and edit this is not schema dependent and can be dealt with a single function. So good if possible, but not a must.

app/routers/content_apis.py
kavitharaju: Could call it "data-manipulation"

app/routers/content_apis.py
```diff
@@ -57,6 +58,43 @@ async def add_contents(request: Request, content: schemas.ContentTypeCreate,
     return {'message': "Content type created successfully",
             "data": data}
 
+@router.delete('/v2/contents',response_model=schemas.ContentDeleteResponse,
```
kavitharaju, line 61: All deleteResponses will have same strcuture right? So we need only a common Response schema. Not seperate for languages and Contents

app/schema/schema_auth.py
```diff
@@ -17,6 +17,7 @@ class ResourceType(str, Enum):
     JOBS = "Background-jobs"
     MEDIA = "Media contents"
     FILE = "File Manipulation Ops"
+    DATAMANIPULATION = "Manipulate Data"
```
kavitharaju, line 20: The description could explain it more. "Make data alterations in the database for special scenarios"

app/schema/schemas.py
```diff
@@ -129,6 +130,51 @@ class Config:
             }
         }
 
+class LanguageDeleteResponse(BaseModel):
+    """Language delete response"""
+    message:str
```
kavitharaju, line 135: Is it only message? In the router function I see you pass the row from deleted table also as "data". Could include that too so that user gets the information of what item was deleted. At least the id and json field, inside the outer "data" field

app/schema/schemas.py
```diff
@@ -129,6 +130,51 @@ class Config:
...
+        '''display example value in API documentation'''
+        schema_extra = {
+            "example": {
+                "itemId": 40
+            }
+        }
+
+class DataRestoreResponse(BaseModel):
+    """Content delete response"""
+    message:str
```
kavitharaju, line 161: here also we can have a "data" field with the content that was restored.

app/schema/schemas.py
```diff
@@ -129,6 +130,51 @@ class Config:
...
+
+class ContentDeleteResponse(BaseModel):
+    """Content delete response"""
+    message:str
+
+class ContentIdentity(BaseModel):
```
kavitharaju, line 167: I agree with Siju, that we dont need different schema object for different types, all with just an itemId. Having one is enough. But what is even better is to not use a body param. Use a query param, itemId, for all delete APIs.

kavitharaju: And "Kratos" in the doc string should be removed. Kratos is a service we use for user login, auth etc. Not directely conneted to languages or contents.

app/test/test_content_types.py
kavitharaju: yes.

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
     response = client.post(UNIT_URL, headers=headers, json=data)
     assert_input_validation_error(response)
 
+def populate_db(db_ : Session): # pylint: disable=C0116
+    db_.execute('''INSERT INTO content_types(content_type) \
```
kavitharaju, line 101: Better remove this function itslef and use the already defined function. Also we dont follow these paractices
1. writing direct SQL quries if it is something doable with SQL Alchemy. We make use of the db_ session and ability to commit things together, roll back commits etc.
2. We dont use direct database changes in tests at all. We only use the APIs we have defined for them. In this case use the POST APIs to create language or create content

app/test/test_content_types.py
kavitharaju: yes. Same comment for populate db function applies here.

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
...
+    t_id = populate_db(db_=db_)
+    data = {"contentId":t_id}
+    #Registerd User can only delete content type
+    #Delete content without auth
+    headers = {"contentType": "application/json", "accept": "application/json"}
+    response = client.delete(UNIT_URL, headers=headers, json=data)
+    assert response.status_code == 401
+    assert response.json()['error'] == 'Authentication Error'
+
+    #Delete content with auth
```
kavitharaju, line 140: Should be checking in the following way.
- let userX create the language/content with POST call
- try using delete API without auth and make sure 403 and appropriate message is recieved
- try using delete API with auth of userYand make sure 403 and appropriate message is recieved
- try using delete API with auth of vachanAdmin and AgAdmin and make sure 403 and appropriate message is recieved
- try using delete API with auth of userX and make sure it is sucessful

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
...
+                "accept": "application/json",
+                'Authorization': "Bearer"+" "+initial_test_users['APIUser2']['token']
+            }
+    response = client.delete(UNIT_URL, headers=headers, json=data)
+    assert response.status_code == 200
+    assert response.json()['message'] == f"Content with identity {t_id} deleted successfully"
+
+def test_delete_content_id_string():
+    '''positive test case, content id as string'''
+    db_= SessionLocal()
```
kavitharaju, line 151: Why are you using this? Dont make any new sessions or do DB operations in tests. Only use APIs

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
...
+    db_.execute('''DELETE FROM content_types WHERE content_type= 'testdata' ''')
+    db_.commit()
+    t_id = populate_db(db_=db_)
+    testid = str(t_id)
+    data = {"contentId":testid}
+    headers = {"contentType": "application/json",
+                "accept": "application/json",
+                'Authorization': "Bearer"+" "+initial_test_users['APIUser2']['token']
+            }
+    response = client.delete(UNIT_URL, headers=headers, json=data)
```
kavitharaju, line 161: I dont get the logic of this testcase. First you delete the content with SQL statements and then attempting the same with API?

app/test/test_content_types.py
```diff
@@ -92,3 +97,198 @@ def test_post_incorrectvalue_contenttype():
...
+    assert response.status_code == 200
+    assert response.json()['message'] == f"Content with identity {t_id} deleted successfully"
+
+def test_delete_incorrectdatatype():
+    '''the input data object should a json with "contentId" key within it'''
```
kavitharaju, line 166: I dont get the logic of this test case either. Of how you are deleting the contents two times. The doc string says something else.

db/seed_DB.sql
kavitharaju: If COPY and copy means the same(which I think it would), there is no harm in this change. But Shimil, its a good practice to avoid changes that are not required for the task at hand. Even if you make changes while developing you can avoid "add"ing it to commits and pushing to repo.

docker/vachan-ENV/bin/python
```diff
@@ -0,0 +1 @@
+python3
```
kavitharaju, line 1: You should not have added your ENV files!!!
Need to remove these from repo. To avoid these files being shown when you do `git status` or `git add` you should include your env path in the `.git-ignore` file

app/test/test_content_types.py
kavitharaju: I think the way you are writing all these tests need to be revisited and all of them changed

kavitharaju (commented):
Since you merged this code, need to send a new PR with the review changes

.project
shimilgithub: done

app/auth/access_rules.json
shimilgithub: done

app/auth/access_rules.json
shimilgithub: done

app/auth/api-permissions.csv
sijumoncy: I think the  operates on is not None means the resource type should be in CSV , bcz the resource find function will be called only if the resource_type is None .

app/crud/structurals_crud.py
shimilgithub: done

app/deleted_contents.json
shimilgithub: done

app/crud/structurals_crud.py
shimilgithub: What has to be given as the type of del_content?I'm having confusion

kavitharaju: no need to specify type for that argument

app/crud/structurals_crud.py
shimilgithub: restore_content will be deleted.but it has to be returned at the end.So data copied to db_content to return

app/routers/content_apis.py
shimilgithub: done

app/routers/content_apis.py
shimilgithub: done

app/schema/schema_auth.py
shimilgithub: done

app/schema/schemas.py
shimilgithub: changed

app/schema/schemas.py
shimilgithub: changed

db/seed_DB.sql
shimilgithub: changed

app/crud/structurals_crud.py
shimilgithub: done

app/routers/content_apis.py
shimilgithub: removed

app/schema/schemas.py
shimilgithub: Response model for DELETE is merged for all APIs now and named as DeleteResponse. In that case how to keep data for different tables?for contents its "data: ContentType = None" and for languages its "data: LanguageResponse = None"

app/schema/schemas.py
shimilgithub: done

app/schema/schemas.py
sijumoncy: data is in JSON format . right ?

shimilgithub: done

app/schema/schemas.py
shimilgithub: done

shimilgithub: done

app/auth/api-permissions.csv
shimilgithub: > I think the operates on is not None means the resource type should be in CSV , bcz the resource find function will be called only if the resource_type is None .

done

app/crud/structurals_crud.py
shimilgithub: > yeah. db_.commit() should not be here I think

done

app/test/test_content_types.py
shimilgithub: done

app/auth/api-permissions.csv
shimilgithub: > v2/restore is POST or PUT , I think its put because its not a new data creation , restore data is already existed one. clarify with @kavitharaju too

Done

app/test/test_content_types.py
shimilgithub: changed

app/test/test_content_types.py
shimilgithub: changed

app/test/test_content_types.py
shimilgithub: changed

Issue #411, opened 2022-08-12
Token based authentication for Apps


kavitharaju:
Now the Apps are identified based on the header value passed by them in the authentication and access control system. This mechanism can be easily mimicked in any api call. So for a certain resource or operation, we are not able to provide additional access to an app, which we cannot provide for a generic  api-user.

Usecases like this may come, where we have rights to display certain contents on VO, but are not permitted to share it over apis to anyone else. 

Then we not only need to identify the app, but also authenticate it via a token mechanism.

kavitharaju:
Check the new version available for Kratos. See if we can upgrade. 
(do testing on docker, to see anything breaks)

kavitharaju:
For documenting our discussions.
Consider spining up a new Kratos instance, connected to same Kratos DB used by the default one, using
- token expiry time as one year
- accepted userole as "app" alone (in traits)

Need to make sure that a user token obtained from this instance, can be used by the auth module of the vachan-api app to verify the app and read required info eventhough, it is not directly connected to this Kratos instance(only the default Kratos instance which is connected to common Kratos DB)

kavitharaju:
To allow better editing options o the access rules, move them from code base to DB.
May need tables for 
- registered apps
- API-permission maps
- acces rules
and any other as required

sijumoncy:
Auth Refactor Tasks List 

- [x] Authentication Change to DB Based from File Based
- [x] update Seed Db scripts for default values and tables
- [x] Moving permission map, access rules and corresponding data to DB
- [x] Permission map, access rules and other required, default values added in csv to upload through seed script
- [x] Implemented Functions to load required data from DB
- [x] Refactored authentication module for the changes
- [x] Necessary changes in test config and test cases
- [x] Endpoints for App ( Register, Generate Key, Remove key, Delete App, Edit )
- [x] Removed Header based app identification to key through query params. Query params for key added for all endpoints
- [x] kratos instance and schema created for handling app (only working on local, need to fix other ymls) ( using V10.1 )

- [ ] Remove API-App from default user and make email and password from ENV
- [ ] enforce to have role connected to an app name when add a new ROLE in db via role add API (appname , description field in role table)
- [ ] app identification as same as user identification using depend ( Future Option )
- [ ] Fix docker yml is not running - can not connect to second kratos instance (test , docker-compose)
- [ ] Filter App schema remove ( used in get users api for filter list )  use db app name , change role to app as field name

APIs Implementation
[...]

PR #454, opened 2022-11-17
Feature/refactor auth

sijumoncy:
**Basic Refactor of Auth module **

- [x]  Authentication Change to DB Based from File Based
- [x] update Seed Db scripts for default values and tables
- [x] Moving permission map, access rules and corresponding data to DB
- [x] Permission map, access rules and other required, default values added in csv to upload through seed script 
- [x] Implemented Functions to load required data from DB
- [x] Refactored authentication module for the changes
- [x] Necessary changes in test config and test cases
- [x] Endpoints for App ( Register, Generate Key, Remove key, Delete App, Edit )
- [x] Removed Header based app identification to key through query params. Query params for key added for all endpoints
- [x] kratos instance and schema created for handling app (only working on local, need to fix other ymls) ( using V10.1 )

Test is not running because of kratos version and issue in config yml (Working fine in local )


app/auth/auth_app.py
```diff
@@ -0,0 +1,214 @@
...
+                if "phone" in app_update_response["traits"]["contacts"]\
+                else None
+        }
+    elif app_update_req.status_code in (404, 400):
+        raise NotAvailableException(app_update_response["error"]["message"])
+    elif app_update_req.status_code == 500:
+        raise GenericException(app_update_response["error"]["status"])
+    return app_updated_data
+
+DEFAULT_APPS = [
```
kavitharaju, line 162: Is this the place we keep this data? Shouldn't it be at seed DB?
Only password fields need to be read from ENV and filled in, in code, rest can be from seed DB, right?

app/auth/auth_app.py
```diff
@@ -0,0 +1,214 @@
...
+        "email": os.environ.get("VACHANADMIN_APP_EMAIL", "vachanadmin@vachan.testing"),
+        "name": "VachanAdmin",
+        "organization": "BCS",
+        "password": os.environ.get("VACHANADMIN_APP_SECRET", "secret@!@#$"),
+        "contacts": {
+            "email": "None",
+            "phone": "None"
+        }
+    },
+    {
```
kavitharaju: We dont need a separate App for API user. If no app token is provided we can assume its API user.

app/db_models.py
```diff
@@ -432,3 +432,83 @@ class Jobs(Base): # pylint: disable=too-few-public-methods
...
     endTime = Column('end_time', DateTime)
+
+############ Auth Tables ##########
+class Roles(Base): # pylint: disable=too-few-public-methods
+    '''Corresponds to table Roles for auth  in vachan DB(postgres)'''
+    __tablename__ = 'roles'
+
+    roleId = Column('role_id', Integer, primary_key=True)
+    roleName = Column('role_name', String, unique=True, index=True)
+    createdUser = Column('created_user', String)
```
kavitharaju, line 445: It may be good to have role description and also a field associated App, which can be null also

app/routers/auth_api.py
```diff
@@ -155,3 +163,104 @@ async def delete_user(user:schema_auth.UserIdentity,request: Request,#pylint: di
     user_id = user.userid
     delete_identity(user.userid)
     return {"message":f"deleted identity {user_id}"}
+
+# Authentication for Apps -------------------------------------------------------------------
+@router.post('/v2/app/register',response_model=schema_auth.RegisterAppResponse,
+    responses={400: {"model": schemas.ErrorResponse},422: {"model": schemas.ErrorResponse},
+    500: {"model": schemas.ErrorResponse}, 409: {"model": schemas.ErrorResponse}},
+    status_code=201,tags=["App"])
+# @get_auth_access_check_decorator
```
kavitharaju: why is the decorator commented off?

app/routers/auth_api.py
```diff
@@ -155,3 +163,104 @@ async def delete_user(user:schema_auth.UserIdentity,request: Request,#pylint: di
...
+async def generate_key(app_email: str,password: types.SecretStr,
+    request: Request,user_details =Depends(get_user_or_none),#pylint: disable=unused-argument
+    app_key: types.SecretStr = Query(None),#pylint: disable=unused-argument,
+    db_: Session = Depends(get_db)):#pylint: disable=unused-argument
+    '''Generate new Key
+    * app_email and password fiels are mandatory
+    * Successful verifiacation will return a key for app'''
+    log.info('In key Login')
+    log.debug('login:%s',app_email)
+    return login_kratos(app_email,password,from_app=True)
```
kavitharaju, line 197: If we use the same function as for users, then where do we tell that its supposed to use the other Kratos?

app/routers/auth_api.py
```diff
@@ -155,3 +163,104 @@ async def delete_user(user:schema_auth.UserIdentity,request: Request,#pylint: di
...
+
+@router.get('/v2/app/delete-key',response_model=schema_auth.LogoutResponse,
+responses={403: {"model": schemas.ErrorResponse},404:{"model": schemas.ErrorResponse},
+401: {"model": schemas.ErrorResponse}}
+,tags=["App"])
+def delete_key(app_key: types.SecretStr,
```
kavitharaju: Auth decorator is missing

app/routers/content_apis.py
```diff
@@ -24,8 +25,9 @@
     tags=["Contents Types"])
 @get_auth_access_check_decorator
 async def get_contents(request: Request,content_type: str = Query(None, example="bible"),
-     skip: int = Query(0, ge=0),limit: int = Query(100, ge=0),
-     user_details =Depends(get_user_or_none),db_: Session = Depends(get_db)):
+    app_key: types.SecretStr = Query(None),#pylint: disable=unused-argument
+    skip: int = Query(0, ge=0),limit: int = Query(100, ge=0),
+    user_details =Depends(get_user_or_none),db_: Session = Depends(get_db)):
```
kavitharaju, line 30: Wouldn't we need a dependency, app_details = Depends(get_app_or_none) ? Of it is it handled inside the auth decorator>

kavitharaju (commented):
Reviewed only the files with core changes. But lets continue the discussions on this PR.

app/auth/auth_app.py
sijumoncy: The basic data of default apps keep in one of the code base file is because we are using this data in other places ( like test conf) .  The seed Db is only using the name of the default app not other details. we can set the password field as well as the email from the ENV.

app/auth/auth_app.py
sijumoncy: yes the code base is working in that way , no token is provided will be treated as request from API. This is currently only used in test , we can remove this app from the default app list if needed

app/db_models.py
sijumoncy: we already have a relation with app and role . right ? app with default role . do we need the relation 2 way . 
If so does that makes complex database operations
register app :
 role add to table first for a new app
app add to app table and  that role add to default role in app table

app/routers/auth_api.py
sijumoncy: Forgot to remove comment , commented on development

app/routers/content_apis.py
sijumoncy: user details dependency is for the user who is requesting ?

app/db_models.py
sijumoncy: WE can enforce to have role connected to an app name when add a new ROLE in db via role add API

app/auth/auth_app.py
kavitharaju: If you are planning to do this in separate PR make sure you create an issue for this or a task in a main issue

app/schema/schema_auth.py
```diff
@@ -43,15 +42,15 @@ class EditUser(BaseModel):
...
-    BCSDEV = 'BcsDeveloper'
+# class AdminRoles(str, Enum):
+#     '''Admin Roles'''
+#     SUPERADMIN = 'SuperAdmin'
+#     VACHANADMIN = 'VachanAdmin'
+#     AGADMIN = 'AgAdmin'
+#     AGUSER = 'AgUser'
+#     VACHANUSER = 'VachanUser'
+#     APIUSER = 'APIUser'
+#     BCSDEV = 'BcsDeveloper'
```
kavitharaju: We could remove the commented portions. Coming back and cleaning up may not happen always. If we need to reference them we will be keeping the version-2 branch seperate, right? So no risk I guess.

app/schema/schema_auth.py
```diff
@@ -63,7 +62,7 @@ class FilterRoles(str, Enum):
 class UserRole(BaseModel):
     """kratos user role input"""
     userid:str
-    roles:List[AdminRoles]
+    roles:list[str]
```
kavitharaju, line 30: When you add remaining tasks to be done, please add one to ensure we write tests for raising  database errors for invalid roles, apps, permissions etc.

app/schema/schema_auth.py
```diff
@@ -150,3 +149,76 @@ class UserUpdateResponse(BaseModel):
...
+    message:str
+    registered_details:RegistrationAppOut
+    key:str
+
+class RegistrationAppIn(BaseModel):
+    """kratos app registration input"""
+    email:str
+    name:str
+    organization:str
+    password:types.SecretStr
```
kavitharaju, line 142: we are expecting Vachan Admin to store these passwords manually right? And also tokens also, right?

app/routers/auth_api.py
```diff
@@ -36,31 +39,32 @@ async def register(register_details:schema_auth.Registration,request: Request,#p
...
 responses={403: {"model": schemas.ErrorResponse},404:{"model": schemas.ErrorResponse},
 401: {"model": schemas.ErrorResponse}}
 ,tags=["Authentication"])
 def logout(request: Request,user_details =Depends(get_user_or_none),#pylint: disable=unused-argument
+    app_key: types.SecretStr = Query(None),#pylint: disable=unused-argument
     db_: Session = Depends(get_db)):#pylint: disable=unused-argument
     '''Logout
-    * Loging out will end the expiry of a token even if the time period not expired.
-    * Successful login will return a token for user for a time period'''
+    * Loging out will end the expiry of a token even if the time period not expired.'''
```
kavitharaju, line 67: If we are to use this API for revoking an APP token, what would be the expected usage?
- user_token of VA or SA in the auth header
- app_key in the query param
Is this right?
What happens then? Isn't the user gonna get logged out instead of the app?
- If we pass the APP token as the auth header, then how can we ensure it is done by authorized user?
 Do we have another API to expire APP token?

db/csvs/auth_resource_types.csv
```diff
@@ -0,0 +1,19 @@
+resource_type_name,resource_type_description
```
kavitharaju, line 1: I think the file could be named `api_resource_types.csv`

docker/Kratos_config/email-password/kratos-app.yml
```diff
@@ -0,0 +1,88 @@
+version: v0.10.1
+
+dsn: memory
+
+serve:
+  public:
+    base_url: http://kratos:4443/
```
kavitharaju: This and the below one are the container names we discussed about changing.

docker/Kratos_config_App/database.yml
```diff
@@ -0,0 +1,117 @@
+version: "3.7"
+
+services:
+  kratos-migrate:
+    image: oryd/kratos:v0.7.0-alpha.1
+    environment:
+      - DSN=${VACHAN_APP_AUTH_DATABASE:-postgres://kratos:secret@postgresd:5434/kratos?sslmode=disable&max_conns=20&max_idle_conns=4}
```
kavitharaju, line 7: I think it would be worth taking a look at the port numbers of the Kratos postgres DBs. Here I see 5434. Some where else I saw 5435. Better stick to one on all places, quickstart, database and test-dependencies to avoid confusion

docker/Kratos_config_App/email-password/kratos.yml
```diff
@@ -0,0 +1,88 @@
+version: v0.4.6-alpha.1
```
kavitharaju, line 1: Is this version correct?

docker/Kratos_config_App/database.yml
```diff
@@ -0,0 +1,117 @@
+version: "3.7"
```
kavitharaju, line 1: Why do we have both folders `docker/Kratos_config` and `docker/Kratos_config_App` with similar files repeated? Can't we just keep one?

docker/run-tests.yml
```diff
@@ -1,37 +1,39 @@
...
+    environment:
+      - VACHAN_POSTGRES_HOST=vachan-db-test
+      - VACHAN_POSTGRES_USER=${VACHAN_POSTGRES_USER:-postgres}
+      - VACHAN_POSTGRES_PASSWORD=${VACHAN_POSTGRES_PASSWORD:-password}
+      - VACHAN_POSTGRES_DATABASE=${VACHAN_POSTGRES_DATABASE:-vachan_test}
+      - VACHAN_POSTGRES_PORT=5432
+      - VACHAN_KRATOS_PUBLIC_URL=http://kratos:4433/
+      - VACHAN_KRATOS_ADMIN_URL=http://kratos:4434/
+      - VACHAN_KRATOS_APP_PUBLIC_URL=http://kratos:4443/
+      - VACHAN_KRATOS_APP_ADMIN_URL=http://kratos:4444/
```
kavitharaju: Are both set of URLs connecting to same "kratos" container here? Should the app ones be different?

kavitharaju (changes requested):

app/auth/auth_app.py
sijumoncy: https://github.com/Bridgeconn/vachan-api/issues/411 - added it as a task

app/schema/schema_auth.py
sijumoncy: ok, will remove these

app/schema/schema_auth.py
sijumoncy: We are testing the role is the global var of roles always and raise error . Will add this test in coming PR

app/schema/schema_auth.py
sijumoncy: this is the registration input schema for Apps can be accessed from VA app for registering app , need all these fields to create an app in kratos ?

app/routers/auth_api.py
sijumoncy: - This end point to revert user token. and if we pass app key in the header as like a user token , I believe will raise an error for unexpected token type .

- App key in the query params is like all other apis to pass the app token from which request is coming.
- The App key revert have another endpoint currently restricted to acces only from VA app and VA, SA can perform the revert operation as per the MAP and Rules

docker/Kratos_config/email-password/kratos-app.yml
sijumoncy: yes , I will change those, check and Inform

docker/Kratos_config_App/database.yml
sijumoncy: 35 is in the local quickstart.yml.

docker/Kratos_config_App/email-password/kratos.yml
sijumoncy: this need to be changed to latest 10.1 , https://github.com/Bridgeconn/vachan-api/issues/411 mentioned for yml fixes

docker/Kratos_config_App/database.yml
sijumoncy: Kratos_config App folder is the one method that i tried earlier for running seperate instance of kratos , that we discussed , kept this for a reference only , will remove once the current setup start working smooth

docker/run-tests.yml
sijumoncy: https://github.com/Bridgeconn/vachan-api/issues/411 yes will change it to kratos-app. right ?

kavitharaju: yes.

sijumoncy:
- [x] Removed API user from default apps ( added only for tests )
- [x] GQL test disable using pylint command in run test yml
- [x] role table updated with app and description column
- [x] refactored app , user methods
- [x] yml files updated
- [x] filter apps enum removed
- [x] bugs and issue fixed in PR comments
- [x] all test fixed other than 1 test case failing for SW

kavitharaju (approved):
Lets proceed with this as we have our tests working now.
Once all APIs are in place we will be able to test them better with automated tests as well as a round of manual testing

PR #455, opened 2022-11-21
Feature/delete

shimilgithub:


app/db_models.py
```diff
@@ -439,6 +440,7 @@ class DeletedItem(Base): # pylint: disable=too-few-public-methods
 
     itemId = Column('item_id', Integer, primary_key=True,autoincrement=True)
     deletedData = Column('deleted_data', JSON)
+    createdUser = Column('created_user', String)
     deletedUser = Column('deleted_user', String)
```
kavitharaju: Here deletedUser and createdUser looks like having the same meaning and purpose. Need not keep both columns, it will only confuse us later. 
In the DB table you can call it `deleted_user`. Within code you may need to use the name "createdUser" if that is required by the auth module. If so, I suggest you map it like this
```createdUser = Column('deleted_user', String)```

File: app/db_models.py [hunk 2]
```diff
@@ -439,6 +440,7 @@ class DeletedItem(Base): # pylint: disable=too-few-public-methods
...
     deletedTime = Column('deleted_time', DateTime, onupdate=func.now())
```
kavitharaju: Shouldn't this be `oninsert` (or whichever attribute is provided by SQLAlchemy with that meaning)? We dont want the deleted time to be altered later upon some other changes, even though there is no chances for that via API. Lets not take the risk.

app/routers/content_apis.py
```diff
@@ -54,35 +56,36 @@ async def add_contents(request: Request, content: schemas.ContentTypeCreate,
     log.debug('content: %s',content)
     if len(structurals_crud.get_content_types(db_, content.contentType)) > 0:
         raise AlreadyExistsException(f"{content.contentType} already present")
-    data = structurals_crud.create_content_type(db_=db_, content=content)
+    data = structurals_crud.create_content_type(db_=db_, \
+        content=content,user_id=user_details['user_id'])
     return {'message': "Content type created successfully",
-            "data": data}
+            "data": data,"createdUser": user_details['user_id']}
```
kavitharaju: Is there a need to return `user_id` in response object? I dont think we follow that anywhere else in our APIs

app/routers/content_apis.py
```diff
@@ -54,35 +56,36 @@ async def add_contents(request: Request, content: schemas.ContentTypeCreate,
...
     log.debug('content-delete:%s',content_obj)
-    content_id= content_obj.contentId
+    content_id= content_obj.itemId
     dbtable_name = "content_types"
     deleted_content = structurals_crud.delete_content(db_=db_, content=content_obj)
     delcont = structurals_crud.add_deleted_data(db_=db_,del_content= deleted_content,
              user_id=user_details['user_id'],table_name = dbtable_name)
     return {'message': f"Content with identity {content_id} deleted successfully",
-            "data": delcont}
+            "data": delcont,"deletedUserData": user_details['user_id']}
```
kavitharaju: Here also, is there a reason why you have added user_id in response?

app/auth/access_rules.json
```diff
@@ -1,13 +1,13 @@
 {
     "meta-content":{
-        "create": ["registeredUser"],
+        "create": ["registeredUser","SuperAdmin"],
```
sijumoncy: SuperAdmin not needed in "create" beacuse of superAdmin is comes in registered user

app/routers/content_apis.py
```diff
@@ -92,6 +95,9 @@ async def restore_content(request: Request, content: schemas.RestoreIdentity,
     log.info('In restore_content')
     log.debug('restore: %s',content)
     data = structurals_crud.restore_data(db_=db_, restored_item=content)
+    data = jsonpickle.encode(data)
+    data = re.sub(r'^.*?}}}, ' ,'{', data)
+    data = json.loads(data)
```
kavitharaju: Could you please explain what you are doing here? 
The `re.sub()` part. Is this how you clean off the sqlAlchemy related items from the json? If so, is this safe to assume that it will always be the first three things that you would want to remove? 
And help me understand what jsonpickle library does here. Wont handling the object as dict and letting the db_model object handle the conversion to postgres's jsonb type be a safer and more standard approach?

app/db_models.py
```diff
@@ -439,6 +440,7 @@ class DeletedItem(Base): # pylint: disable=too-few-public-methods
 
     itemId = Column('item_id', Integer, primary_key=True,autoincrement=True)
     deletedData = Column('deleted_data', JSON)
+    createdUser = Column('created_user', String)
```
sijumoncy: is this created user field is for the deleted resource created user? If so this need to be come inside the json . right ?

app/routers/content_apis.py
sijumoncy, line 62: why you are passing an extra key created user in the create contents api response . is this same data available inside the data get from the crud function

kavitharaju (changes requested):

app/routers/content_apis.py
sijumoncy: right. the user_id is same as created or updated user , which will be part of the CRUD response

app/routers/content_apis.py
kavitharaju: This regex look very unsafe to me. Usually regex matches the longest possible string. So, if "}}}" occurs further down the road in the string, we will loose data.

app/test/test_content_types.py
```diff
@@ -97,84 +96,146 @@ def test_post_incorrectvalue_contenttype():
...
+    #Delete content with other API user
     headers = {"contentType": "application/json",
-                "accept": "application/json",
-                'Authorization': "Bearer"+" "+initial_test_users['APIUser2']['token']
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+initial_test_users['APIUser']['token']
             }
-    response = client.post(UNIT_URL, headers=headers, json=data)
-    assert response.status_code == 201
+    response = client.delete(UNIT_URL, headers=headers, json=data)
```
sijumoncy: use can use a loop if needed instead of all duplicate steps of test for permission denied

app/test/test_content_types.py
```diff
@@ -183,104 +244,180 @@ def test_delete_incorrectdatatype():
...
-    ''' request a non existing content ID, Ensure there is no partial matching'''
-    data = {"contentId":1000}
-    headers = {"contentType": "application/json",
-                "accept": "application/json",
-                'Authorization': "Bearer"+" "+initial_test_users['APIUser2']['token']
-            }
-    response = client.delete(UNIT_URL,headers=headers,json=data)
-    assert response.status_code == 502
-    assert response.json()['error'] == "Database Error"
+def test_delete_seed_data():
```
sijumoncy: I am not sure about we need a test for seed db content can not delete.

-  I did not understand the test case , this test case always raise the exception and does that make the test fails ?
and also the delete is called as Noauth will never perform delete operation . right ?

sijumoncy: you can add a test case for , if the requested content is not available in db with auth

app/test/test_content_types.py
```diff
@@ -183,104 +244,180 @@ def test_delete_incorrectdatatype():
...
-                    'Authorization': "Bearer"+" "+token_admin
-                }
-    response = client.post(RESTORE_URL, headers=headers_admin, json=data)
-    assert response.status_code == 201
-    assert response.json()['message'] == f"Deleted Item with identity \
-    {deleted_item_id} restored successfully"
-    logout_user(token_admin)
+    #only Super Admin can restore deleted data
+    #Creating and Deelting data
+    response = test_delete_default_superadmin()
```
sijumoncy, line 284: why we are calling another test , if we do so the data will not be available after the called test ?

app/test/test_content_types.py
```diff
@@ -183,104 +244,180 @@ def test_delete_incorrectdatatype():
...
+    deleteditem_id = response.json()['data']['itemId']
+    data = {"itemId": deleteditem_id}
```
sijumoncy, line 286: how to ensure that the deleted content is restored. write a get api call for the deleted content to the original table. Its better to check the original table for the content not present after perform delete operation

app/test/test_content_types.py
```diff
@@ -183,104 +244,180 @@ def test_delete_incorrectdatatype():
...
+    test_user_token = response.json()["token"]
+    headers_auth = {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token
+            }
+
+    response = client.put(RESTORE_URL, headers=headers_auth, json=data)
+    assert response.status_code == 201
+    assert response.json()['message'] == \
+    f"Deleted Item with identity {deleteditem_id} restored successfully"
```
sijumoncy, line 322: I have doubt on this restore happened correctly ? plz ensure the test runs as expected logic

app/test/test_content_types.py
```diff
@@ -183,104 +244,180 @@ def test_delete_incorrectdatatype():
...
 def test_restore_item_id_string():
-    '''positive test case, item id as string'''
-    db_= SessionLocal()
-    deleted_item_id = delete_db(db_=db_)
-    item_id = str(deleted_item_id)
-    data = {"itemId": item_id}
+    '''positive test case, passing deleted item id as string'''
+    #only Super Admin can restore deleted data
+    #Creating and Deleting data
+    response = test_delete_default_superadmin()
```
sijumoncy, line 333: @kavitharaju does this work as expected ? calling a test as a function may work . Normally the dB role back happen after completing a test. 
does this treat as a test or a normal function call ?

app/test/test_content_types.py
```diff
@@ -183,104 +244,180 @@ def test_delete_incorrectdatatype():
...
                     "accept": "application/json",
                     'Authorization': "Bearer"+" "+token_admin
                      }
-    response = client.post(RESTORE_URL, headers=headers_admin, json=data)
+
+    response = client.put(RESTORE_URL, headers=headers_auth, json=data)
     assert response.status_code == 201
-    assert response.json()['message'] == "Deleted Item with identity 60 restored successfully"
+    assert response.json()['message'] == \
+    f"Deleted Item with identity {deleteditem_id} restored successfully"
```
sijumoncy, line 357: do we have a json  or something as response of the restored content to check the response is same as expected ?

sijumoncy: add a test case for a non existing content restore, or try restore again the same content after 1 successful restore

app/test/test_languages.py
```diff
@@ -247,6 +249,7 @@ def test_put_languages():
     response = client.put(UNIT_URL, headers=headers_auth, json=data)
     assert response.status_code == 201
     assert response.json()['message'] == "Language edited successfully"
+    print("edited with created user")
```
sijumoncy: remove this if not necessary

app/test/test_languages.py
```diff
@@ -346,90 +349,154 @@ def test_searching():
...
+    data = {"itemId":language_id}
+
+    #Delete without authentication
     headers = {"contentType": "application/json", "accept": "application/json"}
     response = client.delete(UNIT_URL, headers=headers, json=data)
-    assert response.status_code == 422
-    #Delete Language with Auth
-    #Insert New Language
+    assert response.status_code == 401
+    assert response.json()['error'] == 'Authentication Error'
```
sijumoncy, line 364: can use a loop inf needed as same as commented in contents type tests

app/test/test_languages.py
```diff
@@ -346,90 +349,154 @@ def test_searching():
...
+    test_user_token = response.json()["token"]
+    headers_auth = {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token
+            }
+    #Delete language
+    response = client.delete(UNIT_URL, headers=headers_auth, json=data)
     assert response.status_code == 200
-    assert response.json()['message'] == f"Language with identity {lang_id} deleted successfully"
+    assert response.json()['message'] == \
```
sijumoncy, line 413: check for the response object too

sijumoncy: and ensure at least in one test the targeted item is deleted from the original table by a get call

app/test/test_languages.py
```diff
@@ -438,13 +505,186 @@ def test_delete_missingvalue_language_id():
     response = client.delete(UNIT_URL, headers=headers, json=data)
     assert_input_validation_error(response)
 
-def test_delete_notavailable_language_id():
-    ''' request a non existing content ID, Ensure there is not partial matching'''
-    data = {"languageId":1000}
+
+def test_delete_seed_data():
```
sijumoncy: @kavitharaju Do we need a test like this ?

sijumoncy: I feel it like a function not as a test

app/test/test_languages.py
```diff
@@ -438,13 +505,186 @@ def test_delete_missingvalue_language_id():
...
+    #Restore language with VachanAdmin
+    headers = {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+initial_test_users['VachanAdmin']['token']
+            }
+    response = client.put(RESTORE_URL, headers=headers, json=data)
+    assert response.status_code == 403
+    assert response.json()['error'] == 'Permission Denied'
+
+    #Restore language with AgAdmin
```
sijumoncy: add a test case with resource created user if needed

app/test/test_languages.py
```diff
@@ -438,13 +505,186 @@ def test_delete_missingvalue_language_id():
...
+                    'Authorization': "Bearer"+" "+token_admin
+                     }
+
+    response = client.put(RESTORE_URL, headers=headers_auth, json=data)
+    assert response.status_code == 201
+    assert response.json()['message'] == \
+    f"Deleted Item with identity {deleteditem_id} restored successfully"
+    logout_user(token_admin)
+
+def test_restore_incorrectdatatype():
```
sijumoncy, line 615: input data in json type . common post operation inputs data will be in json format. can we make this more specific to our api, like key misssing in the json/ wrong key or invalid values etc.

app/test/test_content_types.py
shimilgithub: responses of these inside called functions are returned.Its coming when checked with print statements.

app/routers/content_apis.py
shimilgithub: JSONPickle is used to serialize python object to json form. This package is used since I am unable to  serialize it with other methods.
Structure of return item is like this - "SQLAlchemy related info }}}, db model values". '}}}' is not coming anywhere else when i tried with different db_models.So data is cleaned and removed sqlalchemy informations using re.sub()

kavitharaju: I still think this is unsafe to use that regex, because we dont know what would be the data we would encounter later.

And my second question was, what is the need for serialization? Can't you work with dictionary after `to_dict` conversion of SQLAlchemy object. Then remove the unwanted values by specifying their names explicitly.

app/test/test_content_types.py
kavitharaju: Good point. 
The response might come, but I do not know if the changes will persist in the DB.
The calling testcase will have a transaction of itsown, but the inner one will create a new transaction and it will be rolled back at the return from inner function. This is what I assume will happen. So the changes happended in the inner function will not be available for the outer function!

app/test/test_languages.py
kavitharaju: I suggested we do this test. But the way it is written doesn't make an effective test case. 
1. No auth header is given. So it will get permission denied anyways. So no relevance to the fact that its a seed data.
2. Raising exception in test cases is not the usual practice. We use "assert" statements for that.
3. The logic of opening the seed DB file and searching for "hi" in it is not really serving any purpose here.

Now that I think about it, even if its a seed DB contents, as per the rules we have, SA will be able to delete it. So better avoid this test. But I explained it in detail so that @shimilgithub understand the problems

app/test/test_content_types.py
shimilgithub: Shall I replace this inner function call with post and delete methods?

kavitharaju: But @shimilgithub you can verify that, if you see DB changes are persisting when you call like this, you can continue using this method. But like Siju said, I also have doubts about it.

kavitharaju:
@shimilgithub  still there are conflicts

kavitharaju:
Have you added a test for the following scenario?
- create a language
- create a source entry (Eg; commentry) in that language
- try to delete that language with all correct details and auth, and make sure it gives error saying that "the value is in use and can't be deleted"

app/test/test_content_types.py
shimilgithub: When I print response of inner function it returns deleted item id and token of deleted user correctly.

kavitharaju: Seeing the "response" is not enough. Need to make sure the DB changes persist

kavitharaju (commented):
The only problem that caught my attention is the introduction of "createdUser" field in response object of Content, Language and even delete API. I dont understand the need for it.

Also the need of column created_user in the deleted_items table.
Rest looks good.

shimilgithub:
> The only problem that caught my attention is the introduction of "createdUser" field in response object of Content, Language and even delete API. I dont understand the need for it.
> 
> Also the need of column created_user in the deleted_items table. Rest looks good.

Removed createdUser from mentioned places

app/test/test_languages.py
```diff
@@ -346,105 +349,331 @@ def test_searching():
...
+        "versionName": "King James Version",
+        "revision": 1,
+        "metaData": {
+            "publishedIn": "1611"
+            }
+        }
+    response = client.post(VERSION_URL, headers=headers_auth, json=version_data)
+    print("AFTER CREATE VERSION",response.json())
+    assert response.status_code == 201
+    assert response.json()['message'] == "Version created successfully"
```
sijumoncy, line 511: I think this createVersion fails later , because KJV is a valid one and probably will be in db , Do we have a chance of getting already Exist error on create ? You can use some not common values because this is just for testing.

app/test/test_languages.py
```diff
@@ -346,105 +349,331 @@ def test_searching():
...
+    #Delete language with item created API User
+    data = {"itemId":language_id}
     headers = {"contentType": "application/json",
                     "accept": "application/json",
                     'Authorization': "Bearer"+" "+initial_test_users['APIUser2']['token']
             }
     response = client.delete(UNIT_URL, headers=headers, json=data)
+    print("AFTER DELETE LANGUAGE",response.json())
+    assert response.status_code == 409
+    assert response.json()['error'] == 'Already Exists'
```
sijumoncy: Why we get a 409 error . this test is for trying to delete a language which used by another table. and this should be a kind of database error ?
409 - is a kind of create something and tellling us this is already there

app/test/test_languages.py
```diff
@@ -346,105 +349,331 @@ def test_searching():
...
+    headers = {"contentType": "application/json",
+                "accept": "application/json",
+                'Authorization': "Bearer"+" "+initial_test_users['APIUser2']['token']
+            }
+    response = client.delete(UNIT_URL,headers=headers,json=data)
+    assert response.status_code == 404
+    assert response.json()['error'] == "Requested Content Not Available"
+
+def test_language_used_by_source():
+    '''  Negativetest case, trying to delete that language which is used to create a source'''
```
sijumoncy, line 477: Do we have similar thing for content types , license, version . those are also part of source creation. Please make sure.

sijumoncy (changes requested):

app/test/test_languages.py
shimilgithub: done

kavitharaju (dismissed):

kavitharaju (approved):

sijumoncy (approved):

PR #463, opened 2022-12-08
Feature/delete

shimilgithub:
committed hard delete on licenses and versions

### DIFF
--- a/app/crud/structurals_crud.py
+++ b/app/crud/structurals_crud.py
@@ -45,10 +45,9 @@ def create_content_type(db_: Session, content: schemas.ContentTypeCreate,user_id
 def delete_content(db_: Session, content: schemas.DeleteIdentity):
     '''delete particular content, selected via content id'''
     db_content = db_.query(db_models.ContentType).get(content.itemId)
-    deleted_content = db_content
     db_.delete(db_content)
     #db_.commit()
-    return deleted_content
+    return db_content
 
 def get_languages(db_: Session, language_code = None, language_name = None, search_word=None,
     language_id = None, **kwargs):
@@ -126,48 +125,45 @@ def get_restore_item_id(db_: Session, restore_item_id = None, **kwargs):
     '''Fetches row of deleted item'''
     skip = kwargs.get("skip",0)
     limit = kwargs.get("limit",100)
-    query = db_.query(db_models.ContentType)
+    query = db_.query(db_models.DeletedItem)
     if restore_item_id is not None:
         query = query.filter(db_models.DeletedItem.itemId == restore_item_id)
     return query.offset(skip).limit(limit).all()
 
 def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
     '''Restore deleted record back to the original table'''
-    restore_content = db_.query(db_models.DeletedItem).get(restored_item.itemId)
-    db_content = restore_content
+    db_restore = db_.query(db_models.DeletedItem).get(restored_item.itemId)
+    db_content=db_restore
     json_string = db_content.deletedData
-    db_.delete(restore_content)
-    if db_content.deletedFrom == 'content_types':
+    db_.delete(db_restore)
+    common_field_table_db_model_dict = {"languages":db_models.Language, \
+        "licenses":db_models.License,"versions":db_models.Version}
+    if db_restore.deletedFrom in common_field_table_db_model_dict:
+        db_content = common_field_table_db_model_dict[db_restore.deletedFrom] \
+            (createdUser= json_string['createdUser'],
+            updatedUser= json_string['updatedUser'],updateTime = datetime.now())
+    if db_restore.deletedFrom == 'content_types':
         db_content = db_models.ContentType(contentId = json_string['contentId'],
-        contentType = json_string['contentType'])
-    elif db_content.deletedFrom == 'languages':
+            contentType = json_string['contentType'],
+            createdUser = json_string['createdUser'])
+    elif db_restore.deletedFrom == 'languages':
         db_content = db_models.Language(scriptDirection = json_string['scriptDirection'],
         code= json_string['code'],
         language= json_string['language'],
         languageId = json_string['languageId'],
-        metaData= json_string['metaData'],
-        createdUser= json_string['createdUser'],
-        updatedUser= json_string['updatedUser'],
-        updateTime = datetime.now())
-    elif db_content.deletedFrom == 'licenses':
+        metaData= json_string['metaData'])
+    elif db_restore.deletedFrom == 'licenses':
         db_content = db_models.License(code = json_string['code'],
         name= json_string['name'],
         license= json_string['license'],
         permissions = json_string['permissions'],
-        active = json_string['active'],
-        #metaData= json_string['metaData'],
-        createdUser= json_string['createdUser'],
-        updatedUser= json_string['updatedUser'],
-        updateTime = datetime.now())
-    elif db_content.deletedFrom == 'versions':
+        active = json_string['active'])
+    elif db_restore.deletedFrom == 'versions':
         db_content = db_models.Version(versionAbbreviation = json_string['versionAbbreviation'],
         versionId= json_string['versionId'],
         versionName= json_string['versionName'],
         revision= json_string['revision'],
-        metaData= json_string['metaData'],
-        createdUser= json_string['createdUser'],
-        updatedUser= json_string['updatedUser'],
-        updateTime = datetime.now())
+        metaData= json_string['metaData'])
     db_.add(db_content)
     #db_.commit()
     return db_content
@@ -175,10 +171,9 @@ def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
 def delete_language(db_: Session, lang: schemas.DeleteIdentity):
     '''delete particular language, selected via language id'''
     db_content = db_.query(db_models.Language).get(lang.itemId)
-    deleted_content = db_content
     db_.delete(db_content)
     #db_.commit()
-    return deleted_content
+    return db_content
 
 def get_licenses(db_: Session, license_code = None, license_name = None,
     permission = None, active=True, **kwargs):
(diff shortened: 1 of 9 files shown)


app/crud/structurals_crud.py
```diff
@@ -149,8 +149,26 @@ def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
...
+        updateTime = datetime.now())
+    elif db_content.deletedFrom == 'versions':
+        db_content = db_models.Version(versionAbbreviation = json_string['versionAbbreviation'],
+        versionId= json_string['versionId'],
+        versionName= json_string['versionName'],
+        revision= json_string['revision'],
+        metaData= json_string['metaData'],
+        createdUser= json_string['createdUser'],
+        updatedUser= json_string['updatedUser'],
+        updateTime = datetime.now())
```
kavitharaju: even if we can't avoid this elif ladder, it would be good if we take the common fields outside.
`createdUser`, `updatedUser`, `updatedTime` etc

app/crud/structurals_crud.py
```diff
@@ -218,6 +236,14 @@ def update_license(db_: Session, license_obj: schemas.LicenseEdit, user_id=None)
     # db_.refresh(db_content)
     return db_content
 
+def delete_license(db_: Session, content: schemas.DeleteIdentity):
+    '''delete particular license, selected via license id'''
+    db_content = db_.query(db_models.License).get(content.itemId)
+    deleted_content = db_content
+    db_.delete(db_content)
+    #db_.commit()
+    return deleted_content
```
kavitharaju: I think I have asked this before, but please remind me, why do we need these two variables `deleted_content` and `db_content`? Can't we just use `db_content` itself as return object?

app/routers/content_apis.py
```diff
@@ -237,6 +237,28 @@ async def edit_license(request: Request, license_obj: schemas.LicenseEdit = Body
...
+    user_details =Depends(get_user_or_none), db_: Session = Depends(get_db)):
+    '''Delete License
+    * unique License Id can be used to delete an exisiting identity'''
+    log.info('In delete_licenses')
+    log.debug('license-delete:%s',delete_obj)
+    license_id= delete_obj.itemId
+    dbtable_name = "licenses"
+    if len(structurals_crud.get_license_id(db_, license_id= delete_obj.itemId)) == 0:
+        raise NotAvailableException(f"License id {license_id} not found")
+    deleted_content = structurals_crud.delete_license(db_=db_, content=delete_obj)
```
kavitharaju, line 256: Can we avoid two database access here?
Avoid the first `structurals_crud.get_license_id()`. 
If we try to delete a non existing item I hope database would raise error. If so, we need only one DB access instead of two for deleting.

app/test/test_licenses.py
```diff
@@ -291,3 +295,307 @@ def test_put(): #pylint: disable=too-many-statements
     resp2 = client.get(UNIT_URL)
     assert resp1.status_code == 200
     assert len(resp1.json()) - len(resp2.json()) == 1
+
+def test_delete_default():
+    ''' positive test case, checking for correct return of deleted license ID'''
+    #create new data
+    response = test_post()
```
kavitharaju: I see this test_post() is a long test case, line 95 to 195, which does several positive and negative tests. I dont see any reason to rerun all of that just so we get the output of one response. It will slow down our tests which are already taking too long to run. Better to have a new function(not test function) that does one successful post operation and returns the response

app/test/test_licenses.py
```diff
@@ -291,3 +295,307 @@ def test_put(): #pylint: disable=too-many-statements
...
+            "publishedIn": "1611"
+            }
+        }
+    response = client.post(VERSION_URL, headers=headers_admin, json=version_data)
+    assert response.status_code == 201
+    assert response.json()['message'] == "Version created successfully"
+
+    #Create Source with license
+    source_data = {
+        "contentType": "commentary",
```
kavitharaju, line 441: For creating source and version here there might be functions you can import from those tests. Worth taking a look there. Eg: look at the commentary tests and see how required version and source are created there by importing functions. Helps with handling the URLs, in case we change them. Keeping URLs in minimum places helps.

kavitharaju (changes requested):
Suggestions have been mentioned only in one place. But when making changes, need to alter all similar cases. Also good to revisit the previous APIs where ever required.

app/routers/content_apis.py
shimilgithub: @kavitharaju  If I remove structurals_crud.get_license_id() ,it shows the following error.(Error - 502)

  "error": "Database Error",
  "details": "Class 'builtins.NoneType' is not mapped"

Is that fine?

shimilgithub: Error message doesn't seems to be meaningful. It would be better if we keep both database accesses.

kavitharaju: Agreed. Lets keep it as it is now.

app/crud/structurals_crud.py
```diff
@@ -133,34 +132,48 @@ def get_restore_item_id(db_: Session, restore_item_id = None, **kwargs):
...
+    db_restore = db_.query(db_models.DeletedItem).get(restored_item.itemId)
+    db_content=db_restore
     json_string = db_content.deletedData
-    db_.delete(restore_content)
-    if db_content.deletedFrom == 'content_types':
+    db_.delete(db_restore)
+    deleted_from_list = ['languages','licenses','versions','sources']
+    db_model_list = [db_models.Language,db_models.License,db_models.Version,db_models.Source]
+    for i in range(len(deleted_from_list)): #pylint: disable=consider-using-enumerate
+        if db_restore.deletedFrom == deleted_from_list[i]:
```
kavitharaju: You could perform this loop and if in one simpler step
```
if db_restore.deletedFrom in deleted_from_list:
  ...
```
And naming the variable `deleted_from_list` something similar to `tables_with_common_fields` or somethings would make more sense to what we are doing.

app/test/test_licenses.py
```diff
@@ -89,6 +92,26 @@ def test_get():
     response = client.get(UNIT_URL+"?permission=abcd")
     assert_input_validation_error(response)
 
+def post_data(): #pylint: disable=missing-function-docstring
```
kavitharaju: Its a very good practice to add a docstring explaing what we do in the function. What looks like obvious to us now, may not seem like that later when we read the code. So add a good docstring that explains briefly what we do here.

kavitharaju (changes requested):

app/crud/nlp_crud.py
```diff
@@ -640,8 +640,7 @@ def gloss_forward_reverse_query(db_, word, source_lang, target_lang, total, *arg
             db_models.TranslationMemory.target_language.has(code=target_lang),
             text("soundex(source_token_romanized) = soundex(:word2)").\
                 bindparams(word2=utils.to_eng(word)),
-            text("levenshtein(source_token,:word3) < 4").bindparams(word3=word)
-            ).order_by(text("lev_score"))
+            text("levenshtein(source_token,:word3) < 4").bindparams(word3=word))
```
kavitharaju: May be you made this change upon resolving conflict, but I think the previous is what we need. Need to order the results according to its relevance

app/crud/nlp_crud.py
```diff
@@ -652,8 +651,7 @@ def gloss_forward_reverse_query(db_, word, source_lang, target_lang, total, *arg
             db_models.TranslationMemory.target_language.has(code=source_lang),
             text("soundex(translation_romanized) = soundex(:word2)").\
                 bindparams(word2=utils.to_eng(word)),
-            text("levenshtein(translation,:word3) < 4").bindparams(word3=word)
-            ).order_by(text("lev_score"))
+            text("levenshtein(translation,:word3) < 4").bindparams(word3=word))
```
kavitharaju: Same here. Need order_by

app/crud/nlp_crud.py
```diff
@@ -757,7 +755,7 @@ def auto_translate_token_logic(db_,tokens, sent, source_lang, target_lang):
                 # splits the source sentence into tokens(in draftmeta)
                 # even if there is no transltion suggestion available.
                 draft, meta = nlp_utils.replace_token(sent.sentence, offset,
-                    "", sent.draftMeta, "untranslated",draft=sent.draft)
+                    token, sent.draftMeta, "untranslated",draft=sent.draft)
```
kavitharaju: This also seems to undo some changes we have made on the version-2. Looks like you are not resolving conflicts properly.

Unless, it was a change you made intenstionally for the task at hand, you should not change what is on version-2

app/crud/nlp_tokenization.py
```diff
@@ -203,6 +203,7 @@ def tokenize(db_:Session, src_lang, sent_list, #pylint: disable=too-many-locals
         if "draftMeta" in sent:
             for meta in sent['draftMeta']:
                 if meta[2] == "confirmed":
+                    print("Meta with confirmed", meta)
```
kavitharaju: Again, issue with conflict resolution. Unnecessary print.

kavitharaju (changes requested):
I see lots of changes that are not related to this task. Looks like it happened with conflict resolution or because of not doing it!
image

Need to pull changes from version-2. Make sure you keep the changes happened there, unless you have reasons for changing some in relation to the task at hand

app/crud/nlp_crud.py
shimilgithub: > May be you made this change upon resolving conflict, but I think the previous is what we need. Need to order the results according to its relevance

@kavitharaju These changes are not made by me.These were the changes that happend on latest pull. I didn't committed those on previous PRs.By mistake that happend now. Shall I remove those files which are not modified by me?

kavitharaju: Removing the file means delete them from code base? No!
You should only remove the changes. 

What has happened here is, you are un-doing all the changes that were done on version-2 recently by me. We dont want that. What ever changes were done on version-2 after your previous pull should be there.

kavitharaju (approved):
Looks good to me

PR #470, opened 2022-12-12
access db passwords using spl characters

### LINKED ISSUE #466: Accept DB passwords with special characters
Now if we set the DB password with `@` when building the SQL alchemey connectio URL it fails. Need to url-encode the password to be able to handle this issue.

Leni-Easo:
fixed #466  commit changes on line no: 2 & 14.

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -11,15 +11,16 @@ postgres_database = os.environ.get("VACHAN_POSTGRES_DATABASE", "vachan")
 postgres_password = os.environ.get("VACHAN_POSTGRES_PASSWORD", "secret")
 postgres_port = os.environ.get("VACHAN_POSTGRES_PORT", "5432")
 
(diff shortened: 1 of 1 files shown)


kavitharaju (changes requested):
There is linting error. Please fix that

app/database.py
```diff
@@ -11,15 +11,17 @@
...
+        f"{postgres_host}:{postgres_port}/{postgres_database}"
+#print(SQLALCHEMY_DATABASE_URL)
 
 connection_url = os.environ.get("VACHAN_POSTGRES_CONN", SQLALCHEMY_DATABASE_URL)
 
 engine = create_engine(connection_url, pool_size=10, max_overflow=20)
 SessionLocal = sessionmaker(autocommit=False, autoflush=False, bind=engine)
 
-# meta = MetaData(naming_convention={
+# meta = MetaData(naming_convention={thon
```
kavitharaju, line 24: Its a good practice to not commit changes that are not related to the task at hand. 
But ignoring this for now.

kavitharaju (approved):

PR #471, opened 2022-12-19
committed hard delete on sources

shimilgithub:


app/crud/structurals_crud.py
```diff
@@ -367,6 +376,15 @@ def get_sources(db_: Session,#pylint: disable=too-many-locals,too-many-branches,
             latest_res.append(res_item)
     return latest_res
 
+def get_source_id(db_: Session, source_id = None, **kwargs):
+    '''Fetches row of source id'''
```
kavitharaju: Why define a new get function? Wouldn't it be better to include this provision, to query by id also, into the above `get_sources()` function? To keep things in one place and have polymorphism, rather than defining new functions for each and every usecase.

app/test/test_versions.py
```diff
@@ -410,6 +410,7 @@ def test_delete_version_id_string():
 def test_delete_incorrectdatatype():
     '''negative testcase. Passing input data not in json format'''
     response = test_post_default()
+
     #Deleting created data
```
kavitharaju, line 414: I see small changes like this new line on several files(5 of the total 12 files), which do not have any impact on our task at hand. Its a good practice to not commit these kind of changes to keep the changes minimal and to the point and also avoid later conflicts where it can be avoided.

app/test/test_sources.py
```diff
@@ -294,7 +295,7 @@ def test_post_duplicate():
         "version": "TTT",
         "year": 2020
     }
-    headers = {"contentType": "application/json", "accept": "application/json"}
+    headers = {"contentType": "application/json", "accept": "application/json"} #pylint: disable=unused-variable
```
kavitharaju, line 299: If this is an unused variable, isn't it better to remove it? When dealing with pylint warnings, our first priority should be to comply with what it suggests. We need to disable only in unavoidable scenarios.

kavitharaju: But I see you are running pylint on these test files, and making changes. We were not doing that earlier. I really appreciate the efforts :+1:

app/test/test_sources.py
```diff
@@ -677,22 +687,21 @@ def test_get_source_filter_access_tag():
...
+    headers_auth = {"contentType": "application/json","accept": "application/json"} #pylint: disable=redefined-outer-name
     #app names
-    API = schema_auth.App.API.value
-    AG = schema_auth.App.AG.value
-    VACHAN = schema_auth.App.VACHAN.value
-    VACHANADMIN = schema_auth.AdminRoles.VACHANADMIN.value
+    # API = schema_auth.App.API.value
+    AG = schema_auth.App.AG.value #pylint: disable=invalid-name
+    VACHAN = schema_auth.App.VACHAN.value #pylint: disable=invalid-name
+    VACHANADMIN = schema_auth.AdminRoles.VACHANADMIN.value #pylint: disable=invalid-name
```
kavitharaju, line 705: Same comment here. Try best to comply with their suggestions.

app/test/test_sources.py
```diff
@@ -1042,4 +1051,289 @@ def check_resp_permission(response,check_list):
...
+    headers_sa = {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token
+            }
+
+    response = client.put(RESTORE_URL, headers=headers_sa, json=data)
+    assert response.status_code == 201
+    assert response.json()['message'] == \
+    f"Deleted Item with identity {deleteditem_id} restored successfully"
+    logout_user(test_user_token)
```
kavitharaju: Its worth adding a GET request also for the source here to ensure the restore was indeed proper.

app/test/test_sources.py
```diff
@@ -1042,4 +1051,289 @@ def check_resp_permission(response,check_list):
...
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+initial_test_users['VachanAdmin']['token']
+            }
+    response = client.delete(UNIT_URL, headers=headers, json=data)
+    assert response.status_code == 200
+    assert response.json()['message'] ==\
+         f"Source with identity {source_id} deleted successfully"
+    #Check source is deleted from sources table
+    check_source_name = client.get(UNIT_URL + "?source_name=hi_TTT_1_commentary", \
+        headers=headers_auth)
```
kavitharaju, line 1108: One more check we could include here is, try a _get commentary call_ 
1. Before delete, which should return proper response
2. After the delete, which should not work
3. After the restore, which should work just as case 1 

to be a little cautious since we are not removing the tables.

kavitharaju (changes requested):
Overall this look good and mostly complete. Adding some suggestions and minor changes.

app/test/test_sources.py
shimilgithub: > One more check we could include here is, try a _get commentary call_
> 
> 1. Before delete, which should return proper response
> 2. After the delete, which should not work
> 3. After the restore, which should work just as case 1
> 
> to be a little cautious since we are not removing the tables.

I didn't understand what is meant by "get commentary call".Can you please explain bit more?

kavitharaju (approved):

PR #475, opened 2023-01-06
Added api for GET, POST for Roles

Leni-Easo:
sending PR for issue#457


app/crud/auth_crud.py
```diff
@@ -0,0 +1,38 @@
...
+    role_id = None, **kwargs):
+    '''Fetches rows of role, with pagination and various filters'''
+    skip = kwargs.get("skip",0)
+    limit = kwargs.get("limit",100)
+    query = db_.query(db_models.Roles)
+    if role_name:
+        query = query.filter(func.lower(db_models.Roles.roleName) == role_name.lower())
+    if role_of_app:
+        query = query.filter(func.lower(db_models.Roles.roleOfApp) == role_of_app.lower())
+    if role_description:
```
sijumoncy: There is no meaning in filter a row of table with its description, I don't think its general practice to filter with description. You can change the query param to a search field and try partial or text search for fields like description , app, role

app/routers/auth_api.py
```diff
@@ -269,3 +270,49 @@ async def delete_app(app_id:str,request: Request,#pylint: disable=unused-argumen
...
+    responses={400: {"model": schemas.ErrorResponse},422: {"model": schemas.ErrorResponse},
+    500: {"model": schemas.ErrorResponse}, 409: {"model": schemas.ErrorResponse}},
+    status_code=201,tags=["Roles"])
+@get_auth_access_check_decorator
+async def create_roles(role_details:schema_auth.Roles,request: Request,#pylint: disable=unused-argument
+app_key: types.SecretStr = Query(None),#pylint: disable=unused-argument
+user_details =Depends(get_user_or_none),#pylint: disable=unused-argument
+db_: Session = Depends(get_db)):#pylint: disable=unused-argument
+    '''Roles of apps
+    * role Id, role name fields are mandatory
```
sijumoncy: this is a post request and how roleId will be a field

sijumoncy: Make this description meaningful enough for a user

app/routers/auth_api.py
```diff
@@ -269,3 +270,49 @@ async def delete_app(app_id:str,request: Request,#pylint: disable=unused-argumen
...
+    * the active status column is mandatory'''
+    log.info('In App Registration')
```
sijumoncy: this is a rore section. right ?

app/schema/schema_auth.py
```diff
@@ -187,3 +186,63 @@ def check_phone(cls, val):#pylint: disable=no-self-argument, inconsistent-return
...
+    data: RoleOut
+
+
+class Roles(BaseModel):
+    """kratos roles input"""
+    roleName: str
+    roleOfApp : str
+    roleDescription : str
+
+class RoleIn(BaseModel):
```
sijumoncy: roles and roleIn looks similar . why we need 2 schema ?

sijumoncy (changes requested):

app/crud/auth_crud.py
Leni-Easo: Ok

app/routers/auth_api.py
Leni-Easo: ok, will correct it

app/schema/schema_auth.py
Leni-Easo: will check it

app/auth/api-permissions.csv
```diff
@@ -120,3 +120,6 @@ Endpoints,Method,RequestApp,FilterResults,ResourceType,Permission
 /v2/bibles/{source_name}/books/{book_code}/format/{output_format},GET,VachanAdmin,False,None,read-via-vachanadmin
 /v2/bibles/{source_name}/books/{book_code}/format/{output_format},GET,API-user,False,None,read-via-api
 /v2/files/usfm/to/{output_format},PUT,None,False,File Manipulation Ops,access-via-api
+/v2/access/roles,POST,VachanAdmin,False,access-control,create
+/v2/access/roles,GET,VachanAdmin,False,access-control,read
+/V2/access/roles,PUT,VachanAdmin,False,access-control,edit
```
sijumoncy: Why you added these permission maps here. These are old files and we no longer using this . This kept for just a reference to work on auth for apps. No need to add here

app/routers/auth_api.py
```diff
@@ -273,46 +273,43 @@ async def delete_app(app_id:str,request: Request,#pylint: disable=unused-argumen
...
     data = create_role(db_,role_details, user_id = user_details['user_id'])
     return {'message': "Role created successfully",
         "data": data}
 
-@router.get('/v2/app/roles',
-    response_model=List[schema_auth.RoleReadResponse],
-    response_model_exclude_unset=True,
-    responses={502: {"model": schemas.ErrorResponse},
-    422: {"model": schemas.ErrorResponse}}, status_code=200, tags=["Roles"])
+@router.get('/v2/access/roles',response_model=List[schema_auth.RoleReadResponse],
```
sijumoncy: is it possible to re use the _RoleOut_ schema here in response instead of a new schema _schema_auth.RoleReadResponse_

db/csvs/auth_resource_types.csv
```diff
@@ -16,4 +16,5 @@ editable,editable contents
 translation-project,translation project
 generic-translation, general translation
 file-ops,file operations
-app,"application related operations"
\ No newline at end of file
+app,"application related operations"
+access-control,"application related Roles"
```
sijumoncy: add a more clear description : the access control refer a set of endpoints which are used for the access-control of vachan engine

sijumoncy (changes requested):

app/auth/api-permissions.csv
Leni-Easo: Ok, will remove that

app/routers/auth_api.py
Leni-Easo: both are same but in RoleReadResponse, roleDescription as optional field

db/csvs/auth_resource_types.csv
Leni-Easo: Ok

app/routers/auth_api.py
sijumoncy: why we are putting description as optional field in response . if no description in input will give none value in the response . right ?

Leni-Easo: yes

app/crud/auth_crud.py
```diff
@@ -0,0 +1,43 @@
...
+
+    db_content = db_models.Roles(roleName = role_details.roleName.lower(),
+        roleOfApp = role_details.roleOfApp,
+        roleDescription = role_details.roleDescription,
+        createdUser= user_id,
+        updatedUser=user_id,active=True)
+    db_.add(db_content)
+    # db_.commit()
+    # db_.refresh(db_content)
+    generate_roles()
```
kavitharaju: This generate_roles() is called to refresh the values kept in-memory that are read from the Db right? I notice this function is called 
- before commit or flush, so will the newly added role be present in the read values?
- before auth verification in the decorator, so would it cause issues if called and API actually fails duw to permission error?
@sijumoncy also

kavitharaju (commented):
This PR doesn't have PUT right?

app/crud/auth_crud.py
sijumoncy: yes , it only have post and get

sijumoncy: @kavitharaju  yes ,that's right we need to move this call to some where in the auth module , where we success authenticating and commit . This may need to think a logic. 
I am not tried yet. this may work

- need to this function (may be other functions in other apis ) to rename in a general name  and add to dbcontent
- check for  the function name present in the db content where we do commit (success auth) and call it from there

kavitharaju: sounds good. I like the idea of having a common named function. We can add one more item in the return object of crud along with `db_content` and the others we have now. Need not put it within `db_content` though. 

In the delete APIs we had to do something similar. [here](https://github.com/Bridgeconn/vachan-api/blob/6f989c8cfede11151a6f9f587dc5adfafef1a03f/app/auth/authentication.py#L302). You can think of a better and uniform way to handle these cases alike.

@shimilgithub also

kavitharaju (approved):

kavitharaju:
@sijumoncy your approval is required to merge this

sijumoncy (approved):

Issue #415, opened 2022-09-07
Use calender versioning


kavitharaju:
For sources we release on gitlab, version should be specified by date of release : ShortYear.ShortMonth.ShortDay e.g. 22.9.7.
https://calver.org/

Make required changes on Vachan-api to ensure the sorting, validation, DB data-types used are all are updated to support this.
 
Also make a plan to accommodate other versioning-schemes for adding sources that are released elsewhere.

kavitharaju:
These might help in implementing sorting and maybe in supporting different schemes
```
>>> from distutils.version import LooseVersion
>>> versions = ["1.7.0", "1.7.0.rc0", "1.8.0"]
>>> lv = [LooseVersion(v) for v in versions]
>>> sorted(lv, reverse=True)
[LooseVersion ('1.8.0'), LooseVersion ('1.7.0.rc0'), LooseVersion ('1.7.0')]
```

or 

```
from distutils.version import StrictVersion
versions = ["1.1.2", "1.0.0", "1.3.3", "1.0.12", "1.0.2"]
versions.sort(key=StrictVersion)
```

kavitharaju:
https://stackoverflow.com/questions/59247203/best-way-to-store-versions-in-postgresql this discussion and options mentioned there are worth considering

kavitharaju:
Again noting down for consideration/discussion. Do we need to change our terminologies related to versions in DB

We call "King James Version"  or "Indian Revisied version", `version`
and correspondingly "KJV" and "IRV" `version_code`.
The numbers associated with them, `5` in IRV5 are called `revisions`, and we have been using intergers to represent them.

I feel that this use of terminology is not quite what comes to mind when we hear the term "version", by which we usually mean a "number like version" say `2.0` or `22.9.12`. But on the other hand our usage might be more tied to the Biblical domain.

Do we need to consider a revision of these terminologies? 

May be call "King James Version"  `name`, "KJV" `short_name` and the `KJV_2_en_Bible`, which we use are a unique value to identify the source item internally, a `long_name` or `unique_name`. Then use the word `version` to denote what we call revision now .

Any thoughts? @joelthe1 @rinsingh

joelthe1:
As you suspected, 'version` means something different in the BT world. e.g. 'King James Version' itself is the name of the version and, as is the case here, a lot of the times literally includes the term 'version' in it.

[...]

PR #477, opened 2023-01-13
Feature/calendar version

kavitharaju:
- Implements #415 
- DB changes
    - Renames `version.revision` to `version.version_tag`, `sources.source_name` to `sources.version`, and some others as per suggetsion
    - Uses an array of strings to denote `version_tag` Eg: ["1611","12","13"], ["1"], ["2","0","3","alpha","1"]. This would enable efficient way to match and sort. Also permits both calendar and semantic versioning schemes.
    - Adds a `labels` field in sources, which can have `latest` and other values in it, to override the assumption of highest number always denote latest.
- User interface changes
   - Calls `revision`, `versionTag`. Keeping other names as such `source_name`, `versionAbbreviation`, `versionName` etc as they were.
   - Source name is also computed same way as before. `<langCode>_<versionAbbreviation>_<tag>_<contentType>`. But now tag need not be an integer strictly eg: `en_KJV_1611.12.31_bible`, `gr_UGNT_4_bible`, etc
   - VersionTag is always shown as a string of dot separated values to users, in input and output. They are oblivious to the fact that it is represented as array in DB.

Advantages
- Flexibility: Need not tie us down to only one scheme. Can support different, as we may not have control over all the sources that creates the data. Calendar, semantic or even just a numbering like 1, 2, 3 we have for IRV and UGNT
- Scalability: By not using date datatype or other similar types and just using a generic array we have scope to adapt to future needs easily
- Backward compatibility: If our older content cannot be assigned a date tag and only a number or semantic versioning is known we could still use that.
- Allows pre-releases: By having a separate label to mark a source as latest within a series we can have others in later dates but ensure the published is the one delivered to common requests.


app/crud/structurals_crud.py
```diff
@@ -213,12 +215,47 @@ def get_versions(db_: Session, version_abbr = None, version_name = None, revisio
...
+            raise UnprocessableException("Version tag should be a string of dot separated numbers"+\
+            " of maximum four parts. Eg.: 2022.1.26, 2.0.1, 3.0, 1.2.3.4") from exe
+        array[i] = num
+    return array
+
+def version_array_to_tag(tag_array):
+    '''converts [2022, 1, 11, 0] to "2022.1.11". Used for naming source and response'''
+    tag_str = ""
+    pass_zero = True
+    for item in reversed(tag_array):
```
sijumoncy: Just for a clarification if my versioning is [ Major, Minor, Revision, Build ] eg : 1.7.2.0
[1,7,2,0] will be returned as 1.7.2    ? is there a chance the zero is important to have in response

kavitharaju: Yes. The trailing zeros were ignored in this implementation. 

But this is to be revised as per suggestion from Joel. Then the version can be any dot separated strings. And will have as many fields as the user inputs. We are making it more flexible from our end and giving more control to user(va admin). So, if user put trailing zeros it will be shown as such.

joelthe1 (approved):

PR #478, opened 2023-01-13
Apis for Auth Permission

sijumoncy:
- Get, Post , put apis for auth permission
- No test cases implemented. Tested manually


app/crud/auth_crud.py
```diff
@@ -0,0 +1,44 @@
...
+    skip = kwargs.get("skip",0)
+    limit = kwargs.get("limit",100)
+    query = db_.query(db_models.Permissions)
+    if permission_name:
+        query = query.filter(func.lower(db_models.Permissions.permissionName) ==\
+            permission_name.lower())
+    if search_word:
+        search_pattern = " & ".join(re.findall(r'\w+', search_word))
+        search_pattern += ":*"
+        query = query.filter(text("to_tsvector('simple', permission_name || ' ' ||"+\
```
kavitharaju, line 22: I wonder if searching like this is case sensitive.

app/crud/auth_crud.py
```diff
@@ -0,0 +1,44 @@
...
+        createdUser= user_id,
+        updatedUser=user_id,
+        active=True)
+    db_.add(db_content)
+    return db_content
+
+def update_auth_permission(db_: Session, details, user_id= None):
+    '''update a row to auth permission table'''
+    db_content = db_.query(db_models.Permissions).get(details.permissionId)
+    if details.permissionDescription:
```
kavitharaju, line 45: Is there a reason why we are not allowing editing the permission name? Do we use it anywhere in DB without FK referencing?

app/routers/auth_api.py
```diff
@@ -269,3 +256,78 @@ async def delete_app(app_id:str,request: Request,#pylint: disable=unused-argumen
     log.debug('app-delete:%s',app_id)
     delete_identity(app_id, app=True)
     return {"message":f"deleted app with id : {app_id}"}
+
+# Authentication for App permission ----------------------------------------------
+@router.get('/v2/app/permissions',response_model=List[schema_auth.PermissionOut],
```
kavitharaju: Would "/v2/access/permissions" be more appropriate? This is permission GET, PUT and POST are not specific to an app, right?

app/routers/auth_api.py
```diff
@@ -269,3 +256,78 @@ async def delete_app(app_id:str,request: Request,#pylint: disable=unused-argumen
...
+responses={401: {"model": schemas.ErrorResponse}}
+,tags=["Auth-Permission"])
+@get_auth_access_check_decorator
+async def get_auth_permissions(request: Request,user_details =Depends(get_user_or_none),#pylint: disable=unused-argument
+    app_key: types.SecretStr = Query(None),#pylint: disable=unused-argument,
+    permission_id: int = Query(None, example=100001),
+    permission_name: str = Query(None, example='create'),
+    search_word: str = Query(None, example="operation"),
```
kavitharaju, line 269: Some other filters with which a user would want to query would be,
apps using this permission, endpoints requiring this permission etc.

app/schema/schema_auth.py
```diff
@@ -187,3 +184,51 @@ def check_phone(cls, val):#pylint: disable=no-self-argument, inconsistent-return
...
+class PermissionCreateInput(BaseModel):
+    """auth permission crearte input"""
+    permissionName:str
+    permissionDescription:str = None
+    class Config:
+        '''display example value in API documentation'''
+        schema_extra = {
+            "example": {
+                "permissionName": "create-data",
+                "permissionDescription": "permission to create-data"
```
kavitharaju: "Permission to make POST calls making new entries to server DB" would make better senses to the user and this example will encourage VA to also provide more descriptive descriptions.

kavitharaju (changes requested):

app/crud/auth_crud.py
sijumoncy: No there is no specific reason for that , according to the initial discussion we are not allowing to edit contents like appname, rolename, permissionname etc. 
No, permissions will be used in permission-map and access-rules both are FK.

kavitharaju: Then it would be good to allow editing of name too. I can image that it would take some time for the users to get a hang of how to name things properly. So scenarios where they want to change the names may occur.

app/routers/auth_api.py
sijumoncy: yes , thats right.
Out all endpoints are named in this way v2/app .... 
Thats why I asked in the chat (solution for schema issue) . can we use everything as v2/auth/.... , because all are connected with auth. 
or else we can proceed endpoints like
* app : v2/app/register ....
* roles : v2/app/roles
* permission : v2/access/permission 
* map : v2/auth/permission-map
* accessRules : v2/access/rules

kavitharaju: We have been calling this module "auth" and we are used to it. But actually auth is what Kratos does and this is more of "access-control". 
Roles are not for Apps right? Its for users. 
So, how about we changes the endpoints this way.
- v2/app/register
- v2/access/roles
- v2/access/permissions
- v2/access/permission-map
- v2/access/rules
And add tag "access-control" for all the above APIs except app/register which should be auth itself.

app/routers/auth_api.py
sijumoncy: does that means , appName or endpoint filter return all the permission details that connected via permission map. right ?

kavitharaju: Yes. If you think GET api of permission-map is a better place to do that, I would agree. Just that we need a provision to do such filters somewhere.

app/routers/auth_api.py
sijumoncy: that means all the 15-16 apis comes under single section ?

app/routers/auth_api.py
sijumoncy: yes , that may be the better place to do that

app/routers/auth_api.py
kavitharaju: They are all tightly connected with the concept of "access" right? So its good to link them so and put under a single section.

kavitharaju: If you want you can provide multiple tags, so that there will be smaller groups and one big group.

app/crud/auth_crud.py
sijumoncy: No, This not case sensitive
case-sensitive

kavitharaju:
@sijumoncy , are the changes finished on this PR? Ready to be reviewed again?

sijumoncy:
# New Changes Done
- seed db changes to python script with auto seed on app start if db not exist .

app/crud/auth_crud.py
```diff
@@ -7,6 +7,46 @@
...
+        query = query.filter(db_models.Permissions.permissionId == permission_id)
+    return query.offset(skip).limit(limit).all()
+
+
+def create_auth_permission(db_: Session, details, user_id= None):
+    '''Add a row to auth permission table'''
+    db_content = db_models.Permissions(permissionName = details.permissionName.lower(),
+        permissionDescription = details.permissionDescription,
+        createdUser= user_id,
+        updatedUser=user_id,
```
shimilgithub, line 35: Does updatedUser also need to be assigned to user_id?

sijumoncy: on the initial go , both are same. right . Does that break anything ?

shimilgithub: Does updatedUser means the one who does modification over ceratedUser's data?So will it be None initially?

sijumoncy: yes , right.

kavitharaju (approved):

PR #482, opened 2023-01-18
Feature/delete

shimilgithub:
Following files are modified by me in this PR

-  auth/api-permissions.csv
- auth/authentication.py
- crud/contents_crud.py
- crud/structurals_crud.py
- db_models.py
- routers/content_apis.py
- schema/schema_content.py
- -schema/schemas_nlp.py
- test/test_commentaries.py
- test/test_dictionaries.py


kavitharaju:
@shimilgithub, this PR contains implementation of DELETE apis for commentary and dictionary, right?

app/auth/authentication.py
```diff
@@ -301,18 +301,26 @@ async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-s
             # All no-auth and role based cases checked and appoved if applicable
             if db_:
                 db_.commit()
-                if (method == "DELETE" and "source" in endpoint) or "restore" in endpoint:
+                if (method == "DELETE" and "source" in endpoint):
```
kavitharaju: Don't we need to refresh the dynamic tables lookup after a restore too?

app/auth/authentication.py
```diff
@@ -301,18 +301,26 @@ async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-s
...
                     db_models.dynamicTables = {}
                     db_models.map_all_dynamic_tables(db_= next(get_db()))
+                if (method == "DELETE" and "commentaries" in endpoint):
+                    db_models.map_all_dynamic_tables(db_= next(get_db()))
+                if (method == "DELETE" and "dictionaries" in endpoint):
+                    db_models.map_all_dynamic_tables(db_= next(get_db()))
```
kavitharaju: Try to form one if statement for all required cases rather than repeating the same steps in different blocks. 

BTW, does deleting rows from these tables require re-loading of table names in the dynamic tables object?

app/crud/contents_crud.py
```diff
@@ -6,13 +6,13 @@
...
-import db_models
-from crud import utils
-from crud.nlp_sw_crud import update_job
-from schema import schemas_nlp
-from custom_exceptions import NotAvailableException, TypeException, AlreadyExistsException
+import db_models #pylint: disable=import-error
+from crud import utils  #pylint: disable=import-error
+from crud.nlp_sw_crud import update_job #pylint: disable=import-error
+from schema import schemas_nlp, schema_content #pylint: disable=import-error
+from custom_exceptions import NotAvailableException, TypeException, AlreadyExistsException  #pylint: disable=import-error
```
kavitharaju, line 13: Disabling import error is okay. Otherwise it might show warining on your IDE, but the way we are running pylint these wont cause linter errors. So its not mandatory to disable them. But okay if you do it too

shimilgithub:
> @shimilgithub, this PR contains implementation of DELETE apis for commentary and dictionary, right?
Yes Hard delete for commentary and dictionary

app/crud/contents_crud.py
```diff
@@ -6,13 +6,13 @@
...
 
-def get_commentaries(db_:Session, *args,**kwargs):
+def get_commentaries(db_:Session, *args,**kwargs): #pylint: disable=too-many-locals
```
kavitharaju: Why would there be too many locals error here? I dont see new variables introduced in the changes? 

But for the record,ry best to avoid disabling such warnings. See if we can comply as much as possible.

app/crud/contents_crud.py
```diff
@@ -190,6 +193,27 @@ def update_commentaries(db_: Session, source_name, commentaries,job_id, user_id=
         "output": {"message": "Commentaries updated successfully","data": db_content_out}}
     update_job(db_, job_id, user_id, update_args)
 
+def get_commentary_id(db_: Session,commentary_id = None,source_name =None,table_name=None,**kwargs):
+    '''Fetches row of commentary id'''
```
kavitharaju: Can we avoid creating a new function for this and use the existing get_commentaries() to do the job. I had made this suggestion before and made the change myself if the sources PR. Please refer that and copy that pattern. 

I am saying this because we dont want new new functioned defined for every new atomic need. Then the number of functions will grow too much and we will have trouble keeping track of what all features we already have and not have. We used to do it that in V1. Thats hard to maintain in the long run, with different developers working together.

app/crud/contents_crud.py
```diff
@@ -281,6 +306,27 @@ def update_dictionary_words(db_: Session, source_name, dictionary_words, user_id
         }
     return response
 
+def get_word_id(db_: Session, word_id = None, source_name = None,table_name = None, **kwargs):
+    '''Fetches row of dictionary id'''
```
kavitharaju: Same comment as above. Use get_dictionary() method itself.

app/db_models.py
```diff
@@ -135,12 +134,12 @@ def ref_string(cls): # pylint: disable=E0213
...
     commentary = Column('commentary', String)
-    createdUser = Column('created_user',String)
     active = Column('active', Boolean)
-    # __table_args__ = (
+    createdUser = Column('created_user', String)
+    __table_args__ = (
     #     UniqueConstraint('book_id', 'chapter', 'verse_start', 'verse_end'),
-    #     {'extend_existing': True}
-    #                  )
+        {'extend_existing': True}
```
kavitharaju, line 141: What is the use of this extend_existing ? I am asking because I dont know/remember.

app/schema/schema_content.py
```diff
@@ -363,6 +363,7 @@ class Config:
 
 class CommentaryResponse(BaseModel):
     '''Response object for commentaries'''
+    commentaryId: int
     book : BibleBook
     chapter: int
     verseStart: int = None
```
kavitharaju, line 369: I wonder if we should add createdUser info also here...

kavitharaju (changes requested):

app/schema/schema_content.py
shimilgithub: > I wonder if we should add createdUser info also here...

Authentication is getting passed always without createdUser.

app/db_models.py
shimilgithub: > What is the use of this extend_existing ? I am asking because I dont know/remember.

During delete ,it returns a database error on metadata instance and asked to specify extend_existing=True

app/schema/schema_content.py
kavitharaju: Its not a technical requirement but a bussiness logic need that I am thinking of.
Supose user A wants to delete a row. He tries and gets a permission denied error, because he was not the createdUser. Now he needs to delete it and dont know who should he ask to do it. But its okay for now. returning that info with every row is going to be make the response object too big. We can have it in GraphQL later

kavitharaju:
> Yes Hard delete for commentary and dictionary
That is the first thing you needed to mention in the PR comment. Or in the title. Its okay now, saying this for future use.

app/auth/authentication.py
shimilgithub: > Don't we need to refresh the dynamic tables lookup after a restore too?

Added dynamic table update after  restore

app/crud/contents_crud.py
```diff
@@ -12,17 +12,10 @@
 from schema import schemas_nlp, schema_content #pylint: disable=import-error
 from custom_exceptions import NotAvailableException, TypeException, AlreadyExistsException  #pylint: disable=import-error
 
-def get_commentaries(db_:Session, *args,**kwargs): #pylint: disable=too-many-locals
+def get_commentaries(db_: Session, source_name: str,book_code: str =None,chapter: int =None,#pylint: disable=too-many-arguments
+    verse: int = None,last_verse: int=None,active: bool= True,
```
kavitharaju: Use kwargs, like before and  please dont disable this pylint warning. You can add commentary_id as a new kwarg

app/crud/contents_crud.py
```diff
@@ -33,6 +26,9 @@ def get_commentaries(db_:Session, *args,**kwargs): #pylint: disable=too-many-loc
         query = query.filter(model_cls.book.has(bookCode=book_code.lower()))
     if chapter is not None:
         query = query.filter(model_cls.chapter == chapter)
+    if commentary_id is not None:
+        query = query.filter(model_cls.commentaryId == commentary_id)
+        return query.offset(skip).limit(limit).all()
```
kavitharaju: The pattern we follow is not to run query (.`all()`)) and have return statements in between in these crud functions but do it at the end after applying all possible filters. I think the same pattern will be harmless to use here. Its just for sticking to the same and a cleaner coding style.

app/crud/contents_crud.py
```diff
@@ -231,6 +215,9 @@ def get_dictionary_words(db_:Session, source_name,search_word =None, **kwargs):#
         query = db_.query(model_cls.word)
     else:
         query = db_.query(model_cls)
+    if word_id:
+        query = query.filter(model_cls.wordId == word_id)
+        return query.offset(skip).limit(limit).all()
```
kavitharaju: Same comment as above. Try to run the query and return only at the end

app/routers/content_apis.py
```diff
@@ -690,7 +690,8 @@ async def get_commentary(request: Request,
     book_code: schemas.BookCodePattern=Query(None, example="1ki"),
     chapter: int = Query(None, example=10, ge=-1), verse: int = Query(None, example=1, ge=-1),
     last_verse: int = Query(None, example=3, ge=-1), active: bool = True,
-    skip: int = Query(0, ge=0), limit: int = Query(100, ge=0),
+    skip: int = Query(0, ge=0), limit: int = Query(100, ge=0),\
+    commentary_id: int= Query(None, example=100000),
```
kavitharaju: It is not mandatory to have id in the router, because you added it in crud.

app/crud/contents_crud.py
kavitharaju: Oh, are you doing this to avoid too-many-locals warning?

kavitharaju (changes requested):
As per our chat discussion, shall wait till you get to do this change as well.

> I see you are adding created user field for each row. I understand it is for checking access. How about using the createdUser of the whole source to authenticate instead of doing it row specific. I think that is how we are doing it in PUT APIs.

app/routers/content_apis.py
kavitharaju: But no harm in adding. Just telling so you know.

app/crud/contents_crud.py
shimilgithub: > Oh, are you doing this to avoid too-many-locals warning?

Yes.Shall I pass commentary_id  as kwargs?will it increase local variable count?

kavitharaju: Try that and see. I personally think local-valraibles number is more toleratable than arguments.

app/crud/structurals_crud.py
```diff
@@ -100,14 +100,17 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
 
-def add_deleted_data(db_: Session, del_content, table_name : str = None):
+def add_deleted_data(db_: Session, del_content, table_name : str = None,source_createduser = None):
     '''backup deleted items from any table'''
     json_string = jsonpickle.encode(del_content)#, unpicklable=False
     json_string=json.loads(json_string)
     del json_string['py/object'],json_string['_sa_instance_state']
+    if source_createduser is not None:
+        created_user = source_createduser
+    else:
```
kavitharaju: Dont copy source_created user value here, use the user from the details of request. Suppose userX created the item and then super admin came in and deleted something. when we later examine the DB we will see userX did it, when he didn't. That's incorrect info.

app/test/__init__.py
```diff
@@ -78,6 +78,9 @@ def check_default_get(unit_url, headers, assert_positive_get):
     '''checks for an array of items of particular type'''
     #without auth
     response = client.get(unit_url,headers=headers)
+    print("/////////////response:",response.json())
+    print("/////////////response code:",response.status_code)
+
```
kavitharaju: This function gets called in almost all test files and it would flood our terminal with print statement. Better to remove them

app/auth/authentication.py
```diff
@@ -310,7 +310,7 @@ async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-s
             if check_right(user_details, required_rights, obj, db_):
                 if db_:
                     db_.commit()
-                    if (method == "DELETE" and "source" in endpoint) or "restore" in endpoint:
+                    if (method == "DELETE" and "source" in endpoint):
```
kavitharaju, line 312: Can you double check that restore is indeed included? I see you've resolved the conversation But not see it here

kavitharaju (changes requested):

app/auth/authentication.py
shimilgithub: > Can you double check that restore is indeed included? I see you've resolved the conversation But not see it here

restore test returns 404 error when this endpoint is added.Else it works fine

kavitharaju: I dont get the reason behind 404 error here. How would that happen, if we add an if condition?

Do we have tests to check the following

1. add a source(commentary)
2. Add some content in it
3. delete the source
4. restore the source
5. Acces the added content via GET.

I am afraid this may not work smoothly if we dont refresh the dynamic tables list here.

shimilgithub: > I dont get the reason behind 404 error here. How would that happen, if we add an if condition?
> 
> Do we have tests to check the following
> 
> 1. add a source(commentary)
> 2. Add some content in it
> 3. delete the source
> 4. restore the source
> 5. Acces the added content via GET.
> 
> I am afraid this may not work smoothly if we dont refresh the dynamic tables list here.

these tests are done in test_sources and works fine with and without adding "restore" endpoint in auth. But it raises NotAvailableException(f'{source_name} not found in database in test_commentaries.

I didn't get much clarity on it

kavitharaju: Does point 5 work or throws the error you mentioned? It should work successfully as we restored it. I think that check on "restore" is required for it work.

shimilgithub: > Does point 5 work or throws the error you mentioned? It should work successfully as we restored it. I think that check on "restore" is required for it work.

Yes. point 5 also returns 404 error.

kavitharaju: You mean it throws error in the current implementation, right? ie, without including the if option for restore in endpoint. 
Then need to fix it.

shimilgithub: > You mean it throws error in the current implementation, right? ie, without including the if option for restore in endpoint. Then need to fix it.

It fails in both cases.with and without restore endpoint

kavitharaju: Well then it needs fixing then.

app/auth/authentication.py
```diff
@@ -301,9 +301,8 @@ async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-s
             # All no-auth and role based cases checked and appoved if applicable
             if db_:
                 db_.commit()
-                if (method == "DELETE" and "source" in endpoint):
-                    db_models.dynamicTables = {}
-                    db_models.map_all_dynamic_tables(db_= next(get_db()))
+                # if (method == "DELETE" and "source" in endpoint) or "restore" in endpoint:
```
kavitharaju, line 304: I see you've removed this code, which I don't fully understand. but Shall approve and merge for now. This may need reversal before we finish other DELETEs.

app/crud/contents_crud.py
```diff
@@ -316,14 +317,14 @@ def delete_dictionary(db_: Session, delitem: schema_content.DeleteIdentity,table
     model_cls = table_name
     query = db_.query(model_cls)
     db_content = query.filter(model_cls.wordId == delitem.itemId).first()
-    db_.flush()
-    db_.delete(db_content)
-    #db_.commit()
+    dictionary_createduser = source_db_content.updatedUser
```
kavitharaju: Why are you using `source_db_content.updatedUser` instead of the `user_id` available in this API call?
My concern is, wouldn't this value be the user_id of the user for did the previous update, instead of the current user?

app/crud/structurals_crud.py
```diff
@@ -100,17 +100,21 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
+def add_deleted_data(db_: Session, del_content, table_name : str = None,source_createduser = None,#pylint: disable=too-many-arguments
+    item_createduser=None,user_details=None):
     '''backup deleted items from any table'''
     json_string = jsonpickle.encode(del_content)#, unpicklable=False
     json_string=json.loads(json_string)
     del json_string['py/object'],json_string['_sa_instance_state']
-    if source_createduser is not None:
-        created_user = source_createduser
+    if "SuperAdmin" in user_details['user_roles']:
+        deleted_user = user_details['user_id']
```
kavitharaju: Why would you need a special check for SuperAdmin. Also I dont see the need for this `if-else` checks at all. I did so in my code, because I didn't want to change the portions you did and also didn't wanted to break the existed tests you had. But I think this should be
```
deleted_user = user_details['user_id']
```
without any kind of checks. Then you can get rid of the arguments `source_createduser` and `item_createduser` altogether. If you have any counter arguments, please do share, though.

kavitharaju (commented):
I think there are changes required. But I am open to hearing your side. So please clarify or proceed with the changes

app/auth/authentication.py
shimilgithub: ok

app/crud/structurals_crud.py
shimilgithub: > Why would you need a special check for SuperAdmin. Also I dont see the need for this `if-else` checks at all. I did so in my code, because I didn't want to change the portions you did and also didn't wanted to break the existed tests you had. But I think this should be
> 
> ```
> deleted_user = user_details['user_id']
> ```
> 
> without any kind of checks. Then you can get rid of the arguments `source_createduser` and `item_createduser` altogether. If you have any counter arguments, please do share, though.

If I set "`deleted_user = user_details['user_id']"` and " `dictionary_createduser = user_id`", test delete returns 200 for all the users and noauth.Following is happening when I execute inside router
eg : Delete Dictionary(no auth)
1.delete dictionary function executed
2.Add deleted item executed
3. user_details['user_id']: None   and  resp_obj.createdUser: None.  
4. This validates the check_right in authentication like this.
   ```
if "resourceCreatedUser" in required_rights  and \
                user_details['user_id'] == resp_obj.createdUser:
                print("user_details['user_id']:",user_details['user_id'])
                print("resp_obj.createdUser:",resp_obj.createdUser)
                valid = True
```
                
 Same thing happens for other unauthorized users also

Comment (kavitharaju) [reply app/crud/structurals_crud.py]: 1. You should not overwirte the createdUser value in the sources. That should be updated only once when the dictionary itself is created.
2. The authentication should be happening using the createdUser of sources table and not that of deletedItems table.
     For that, I think you can change the return object of this function this way
```
return {
        'db_content':db_content,
        'source_content':source_row
        }
```
:warning: Do verify the proper syntax, variable names etc and use this method appropriately here. I am just sharing the overall logic.

Comment (shimilgithub) [reply app/crud/structurals_crud.py]: > 1. You should not overwirte the createdUser value in the sources. That should be updated only once when the dictionary itself is created.
> 2. The authentication should be happening using the createdUser of sources table and not that of deletedItems table.
>    For that, I think you can change the return object of this function this way
> 
> ```
> return {
>         'db_content':db_content,
>         'source_content':source_row
>         }
> ```
> 
> warning Do verify the proper syntax, variable names etc and use this method appropriately here. I am just sharing the overall logic.

Do I need to split "`add_deleted_item`"  function into two? One for content types,languages etc, which does not have `source_content` to return and other one for dictionaries,commentaries etc. with above mentioned return object?

Comment (kavitharaju) [reply app/crud/structurals_crud.py]: Oh... try doing it with an `if`, if possible. If its complicated, go for separate functions :+1:

File: app/crud/structurals_crud.py [hunk 19]
```diff
@@ -100,25 +100,29 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
+
+    if source is not None:
+        response =  {
+            'db_content':db_content,
+            'source_content': source
+                }
+    else:
+        response =  {
+        'db_content':db_content,
+        'source_content':del_content
```
kavitharaju, line 123: I wonder how passing language or version object as `source_content` will be treated in the auth module. But as tests work I shall approve this.

app/test/test_agmt_translation2.py
```diff
@@ -207,60 +208,3 @@ def test_draft_update_negative():
         headers=headers_auth, json=put_data2)
     assert resp.status_code == 422
     assert resp.json()['details'] == "Incorrect metadata:Target segment (0, 10), is improper!"
-
-
-def test_empty_draft_initalization():
```
kavitharaju: Did you do conflict resolution properly? Why is this test deleted?

app/test/test_agmt_translation2.py
```diff
@@ -27,11 +32,7 @@
     {"sentenceId": 103,
      "sentence": "One day the fox wished he had tiger's striped coat."},
     {"sentenceId": 104,
-     "sentence": "The good friend tiger lent it to him."},
-    {"sentenceId": 105,
-     "sentence": "Tharjima illatha thettu vaakkukal mathram."},
-    {"sentenceId": 106,
-     "sentence": "Oru jungle allathe mattellam tharjima illatha thettu vaakkukal mathram."}
```
kavitharaju: Why are these changes deleted?

app/test/test_translation_suggestions.py
```diff
@@ -261,45 +261,7 @@ def test_learn_n_suggest():
     assert "ഒരു ടെസ്റ്റ് കേസ്." in draft
     assert "ടെസ്റ്റ് കേസ് ടെസ്റ്റ് ചെയ്തു" in draft or "ടെസ്റ്റ് കേസ് ടെസ്റ്റഡ്" in draft
     assert "ടെവെലപ്പര്‍" in draft
-    assert "ഇത് ആണ് ടെസ്റ്റ്" in draft
+    assert "ഇത് ആണ്  sad story of a poor ടെസ്റ്റ് " in draft
 
-def test_bug_fix():
-    '''testing bug fix for issue #412'''
```
kavitharaju: More changes that are being deleted upon conflict resolution.

app/crud/nlp_crud.py
```diff
@@ -640,8 +640,7 @@ def gloss_forward_reverse_query(db_, word, source_lang, target_lang, total, *arg
             db_models.TranslationMemory.target_language.has(code=target_lang),
             text("soundex(source_token_romanized) = soundex(:word2)").\
                 bindparams(word2=utils.to_eng(word)),
-            text("levenshtein(source_token,:word3) < 4").bindparams(word3=word)
-            ).order_by(text("lev_score"))
+            text("levenshtein(source_token,:word3) < 4").bindparams(word3=word))
```
kavitharaju: Again undoing changes I had made.

kavitharaju (changes requested):
Looks like you have messed up the conflict resolution. If I merge this, upon merging to the version-2 you will undo several of the bug fixes I have done there. These changes are required. Please dont remove them.

shimilgithub:
> Looks like you have messed up the conflict resolution. If I merge this, upon merging to the version-2 you will undo several of the bug fixes I have done there. These changes are required. Please don't remove them.

This was happened on latest pull which was done before the first commit of this thread.Some tests were not working(not the part I modified) without adding this files in git.And it was not showing any conflict. Thats why I mentioned files which are modified by me in comment.

Do I need to pull again from version-2 and send PR?

joelthe1:
Don't have full context here but I'd recommend doing a pull with rebase to
overcome such headaches partially.

kavitharaju (approved):

PR #483, opened 2023-01-20
Added PUT API for Roles

Leni-Easo:
sending PR for issue#457


app/routers/auth_api.py
```diff
@@ -269,3 +271,21 @@ async def delete_app(app_id:str,request: Request,#pylint: disable=unused-argumen
     log.debug('app-delete:%s',app_id)
     delete_identity(app_id, app=True)
     return {"message":f"deleted app with id : {app_id}"}
+#################################Roles########################################
+
+@router.put('/v2/access/roles',response_model=schema_auth.RoleUpdateResponse,
```
sijumoncy: is this same response schema that used as response schema in post, get api PR. Is not try to reuse the same schema

app/routers/auth_api.py
```diff
@@ -269,3 +271,21 @@ async def delete_app(app_id:str,request: Request,#pylint: disable=unused-argumen
...
+responses={400: {"model": schemas.ErrorResponse},422: {"model": schemas.ErrorResponse},
+    500: {"model": schemas.ErrorResponse}, 404: {"model": schemas.ErrorResponse}},
+    status_code=201,tags=["Access-control"])
+@get_auth_access_check_decorator
+async def update_roles(role_details:schema_auth.RoleIn,request: Request,#pylint: disable=unused-argument
+    app_key: types.SecretStr = Query(None),#pylint: disable=unused-argument
+    user_details =Depends(get_user_or_none),db_: Session = Depends(get_db)):#pylint: disable=unused-argument
+    '''Changes one or more fields of roles
+    * Roles should be registered one
```
sijumoncy: I did not get , what mean by _Roles should registered one_. Try to add docs with more clarification

app/schema/schema_auth.py
```diff
@@ -187,3 +187,35 @@ def check_phone(cls, val):#pylint: disable=no-self-argument, inconsistent-return
...
+        }
+
+class RoleIn(BaseModel):
+    """kratos roles update input"""
+    roleId :int
+    roleName: str
+    roleOfApp : str
+    roleDescription : str 
+    
+class RoleUpdateResponse(BaseModel):
```
sijumoncy: create role also have same type of  schema . right ?

sijumoncy (changes requested):

app/routers/auth_api.py
Leni-Easo: I used different schema for PUT not the one use in POST & GET

app/routers/auth_api.py
Leni-Easo: In this I mean try to avoid the repetition of the roles. will correct it.

app/schema/schema_auth.py
Leni-Easo: yes

kavitharaju (approved):

sijumoncy (approved):

PR #496, opened 2023-02-03
hard delete on bible videos

shimilgithub:


app/test/test_bible_videos.py
```diff
@@ -545,4 +547,372 @@ def test_get_access_with_user_roles_and_apps():
...
+    response = login(as_data)
+    assert response.json()['message'] == "Login Succesfull"
+    test_user_token = response.json()["token"]
+    headers_sa= {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token
+            }
+
+    data = {
+      "itemId":200000,
```
sijumoncy, line 765: Just a comment about a thought. No need to fix
its better to use an id number less than our db starting value than a too large value.

sijumoncy, line 765: Just a comment about a thought. No need to fix
its better to use an id number less than our db starting value than a too large value.

app/test/test_bible_videos.py
```diff
@@ -545,4 +547,372 @@ def test_get_access_with_user_roles_and_apps():
...
+    #Restoring data
+    #Restore without authentication
+    headers = {"contentType": "application/json", "accept": "application/json"}#pylint: disable=redefined-outer-name
+    response = client.put(RESTORE_URL, headers=headers, json=data)
+    assert response.status_code == 401
+    assert response.json()['error'] == 'Authentication Error'
+
+
+    #Restore content with other API user,VachanAdmin,AgAdmin,AgUser,VachanUser,BcsDev and APIUSer2
+    for user in ['APIUser','VachanAdmin','AgAdmin','AgUser','VachanUser','BcsDev','APIUser2']:
```
sijumoncy, line 791: API user and API user 2 are same right ? . same access level ?.

sijumoncy (approved):

PR #498, opened 2023-02-07
hard delete on bible books

shimilgithub:


app/crud/contents_crud.py
```diff
@@ -882,6 +884,29 @@ def get_available_bible_books(db_, source_name, book_code=None, content_type=Non
...
+    source_name=None,user_id=None):
+    '''delete particular item from bible, selected via sourcename and bible content id'''
+    source_db_content = db_.query(db_models.Source).filter(
+        db_models.Source.sourceName == source_name).first()
+    model_cls  = db_models.dynamicTables[source_name]
+    model_cls2 =  db_models.dynamicTables[source_name+'_cleaned']
+    query = db_.query(model_cls)
+    query2 = db_.query(model_cls2)
+    db_content = query.filter(model_cls.bookContentId == delitem.itemId).first()
+    db_content2 = query2.filter(model_cls.book_id == model_cls2.book_id).first()
```
sijumoncy: is this model_cls.bookId or db_content.bookId ? 
model class is the dynamic table contains multiple books..... ?
line 897 is to fetch all the rows of the cleaned bible of the book same as the single row of orginal usfm  deleting from dynamic bible table ?

app/crud/structurals_crud.py
```diff
@@ -150,7 +150,22 @@ def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
...
+            model_cls = db_models.dynamicTables[source.sourceName]
+            source_table = db_restore.deletedFrom.replace("_cleaned", "")
+            source = get_sources(db_, table_name=source_table)[0]
+            model_cls2 = db_models.dynamicTables[source.sourceName+'_cleaned']
+            cleanedtable_itemid = db_content.itemId + 1
+            db_restore2 = db_.query(db_models.DeletedItem).get(cleanedtable_itemid)
+            db_content2=db_restore2
+            json_string2 = db_content2.deletedData
+            db_content2 = utils.convert_dict_to_sqlalchemy(json_string2, model_cls2)
+            db_.add(db_content2)
```
sijumoncy, line 168: I did not get the logic completely on restore orginal and cleaned bible tables .
I saw one db content is adding to the db . what about the second table ?
Do we using a logic of restore bible and cleaned with the usfm original data from the deleted table

app/routers/content_apis.py
```diff
@@ -584,6 +584,36 @@ async def get_available_bible_book(request: Request,
...
+    cleaned_tablename = dbtable_name+'_cleaned'
+    get_bible_response = contents_crud.get_available_bible_books(db_, source_name=source_name, \
+        biblecontent_id= delete_obj.itemId)
+    if len(get_bible_response['db_content']) == 0:
+        raise NotAvailableException(f"Bible Book with id {biblecontent_id} not found")
+    deleted_content = contents_crud.delete_bible_book(db_=db_,delitem=delete_obj,\
+        source_name=source_name,user_id=user_details['user_id'])
+    delcont = structurals_crud.add_deleted_data(db_=db_,del_content= deleted_content['db_content'],
+        table_name= dbtable_name,source=deleted_content['source_content'],user_details=user_details)
+    del_clean=structurals_crud.add_deleted_data(db_=db_,del_content=deleted_content['db_content2'],#pylint: disable=unused-variable
```
sijumoncy, line 612: is this really required to store cleaned data too in the db ?

sijumoncy (changes requested):

app/crud/structurals_crud.py
shimilgithub: When we delete a row from the dynamic table of bible,associated cleaned table data also gets deleted along with it and both of them will be added to deleted_items table.
Hence during restore,same happens.That is, cleaned table data also will be restored when we restore the bibles table,

shimilgithub: > I did not get the logic completely on restore orginal and cleaned bible tables . I saw one db content is adding to the db . what about the second table ? Do we using a logic of restore bible and cleaned with the usfm original data from the deleted table

When we delete from the dynamic table of bible,associated cleaned table also getting deleted and both of them will be added to deleted items table.

Hence during restore also same happens.Cleaned table will be restored along with associated bibles table restore

app/routers/content_apis.py
shimilgithub: > is this really required to store cleaned data too in the db ?

This would be the better way since we are accumulating all the deleted contents in a single table.That would be easy for future reference also.
It won't take extra space since we are just moving item from one table to another.
Also when we restore bibles,this cleaned table will automatically restored back which won't take much headache.

Instead of adding cleaned table item into deleted items table and restore,we can delete it and perform insert instead of restore.That  again need to generate ref_id ,split data and convert to usfm format etc. which I felt unnecessary since we already have that  info.

This is my thought.Please let me know if its not feasible.

app/crud/contents_crud.py
shimilgithub: > is this model_cls.bookId or db_content.bookId ? model class is the dynamic table contains multiple books..... ? line 897 is to fetch all the rows of the cleaned bible of the book same as the single row of orginal usfm deleting from dynamic bible table ?

Yes. Thats right. Resolved issue

app/routers/content_apis.py
sijumoncy: ok. Moving cleaned tables from cleaned to delete will effect same storage. make sense

app/crud/structurals_crud.py
sijumoncy: sorry for asking this again .
on delete I found 2 delete performing for db_content, db_content2 for bible and cleaned respectivily. 
on restore there is only 1 item db_content2 is adding back to db. does this single add will restore both bible and cleaned table in a single restore call for one bible book

sijumoncy: one more doubt , what will happen if I perform the following operation
 - source added
 - bible inserted ( create bible and cleaned )
 - delete bible  (remove bible and cleaned )
 - delete source 
 - try restore bible (bible + cleaned )
 
 does this flow cause any error ?

shimilgithub: > sorry for asking this again . on delete I found 2 delete performing for db_content, db_content2 for bible and cleaned respectivily. on restore there is only 1 item db_content2 is adding back to db. does this single add will restore both bible and cleaned table in a single restore call for one bible book

There are two restore happening only in case of bibles. Line 167 is to restore cleaned table and line 171 is to restore bibles.

sijumoncy: > > sorry for asking this again . on delete I found 2 delete performing for db_content, db_content2 for bible and cleaned respectivily. on restore there is only 1 item db_content2 is adding back to db. does this single add will restore both bible and cleaned table in a single restore call for one bible book
> 
> There are two restore happening only in case of bibles. Line 167 is to restore cleaned table and line 171 is to restore bibles.

Ok got it .

shimilgithub: > one more doubt , what will happen if I perform the following operation
> 
> * source added
> * bible inserted ( create bible and cleaned )
> * delete bible  (remove bible and cleaned )
> * delete source
> * try restore bible (bible + cleaned )
> 
> does this flow cause any error ?

This will throw an error because we are trying to fetch source details at line 152

sijumoncy: Do we handled it already ? 
for the other endpoints also ?

shimilgithub: > Do we handled it already ? for the other endpoints also ?

In some endpoints like language,content_types we are perform direct delete and restore.
And the others which are associated with source,we are fetching the source name first from the sources table. If source is not available,it will throw error

sijumoncy (approved):

Issue #405, opened 2022-07-29
Implement hard delete


kavitharaju:
Now we  have the feature of "soft delete" for all the contents in our DB. We are not providing users ability to delete something from Db, but only allowing them to de-activate the contents by setting the status flag to False.

Based on this article, [Soft Deletion Probably Isn't Worth It](https://brandur.org/soft-deletion), we could consider implementing a hard delete on our DB and keeping a **deleted_items_table** to be able to keep track of older contents and be able to manually recover things if required.

Another disadvantage we have in our current system of using soft delete, which was not mentioned in the article is this:

Suppose we have an item created by a user, which was wrong and was deactivated. Later if another entry needs to be made, with same value for a field with unique constraint(say name), then we only have the option of re-activating(editing) the older entry. This might become a problem in a scenario where access rights on the item clashes with our interests. For example, if a project was created by the name "Hindi_BCS_translation" and was de-activated. Suppose we later have a project which needs to be named the same, and the user handling the new project is a different person, he/she will not have edit rights on the first created-and-deactivated project. That name kind of gets taken permanently. This kind of problem can arise with all types of contents like projects, token-translations, dictionary entries, Bible and other contents etc etc.

kavitharaju:
Tasks breakdown
- [x] DELETE v2/contents
- [x]  DELETE v2/languages
- [x] DELETE v2/licenses
- [x] DELETE v2/versions
- [x] DELETE v2/sources
- [x] DELETE v2/commentaries/{source}
- [x] DELETE v2/dictionaries/{source}
- [x] DELETE v2/infographics/{source}
- [x] DELETE v2/biblevideos/{source}
- [x] DELETE v2/bibles/{source}/books
- [x] DELETE v2/bibles/{source}/audios
- [x] DELETE v2/autographa/projects
- [x] DELETE v2/autographa/project/user 
- [x] DELETE v2/autographa/project/sentences
- [x] DELETE v2/nlp/gloss
- [x] DELETE v2/lookup/stopwords/{lang}
- [x] PUT v2/restore
- [ ] DELETE v2/deleted-items

shimilgithub:
> Tasks breakdown
> 
> * [x]  DELETE v2/contents       
> * [x]   DELETE v2/languages
> * [x]  DELETE v2/licenses
> * [x]  DELETE v2/versions
> * [ ]  DELETE v2/sources
> * [ ]  DELETE v2/commentaries/{source}
[...]

PR #500, opened 2023-02-08
hard delete on bible audio

shimilgithub:
Issue #405

### DIFF
--- a/app/crud/contents_crud.py
+++ b/app/crud/contents_crud.py
@@ -799,6 +799,24 @@ def update_bible_audios(db_: Session, source_name, audios, user_id=None):
         }
     return response
 
+def delete_bible_audio(db_: Session, delitem: schema_content.DeleteIdentity,\
+    source_name=None,user_id=None):
+    '''delete particular item from bible audio, selected via sourcename and bible audio id'''
+    source_db_content = db_.query(db_models.Source).filter(
+        db_models.Source.sourceName == source_name).first()
+    model_cls =  db_models.dynamicTables[source_name+'_audio']
+    query = db_.query(model_cls)
+    db_content = query.filter(model_cls.audioId == delitem.itemId).first()
+    db_.flush()
+    db_.delete(db_content)
+    #db_.commit()
+    source_db_content.updatedUser = user_id
+    response = {
+        'db_content'    :db_content,
+        'source_content':source_db_content
+        }
+    return response
+
 def get_bible_versification(db_, source_name):
     '''select the reference list from bible_cleaned table'''
     model_cls = db_models.dynamicTables[source_name+"_cleaned"]
@@ -838,23 +856,25 @@ def get_bible_versification(db_, source_name):
         }
     return response
 
-def get_available_bible_books(db_, source_name,book_code=None, content_type=None,
+def get_available_bible_books(db_, source_name,book_code=None, content_type=None,#pylint: disable=too-many-locals
     biblecontent_id=None, **kwargs):
     '''fetches the contents of .._bible table based of provided source_name and other options'''
     active = kwargs.get("active",True)
     skip = kwargs.get("skip",0)
     limit = kwargs.get("limit",100)
-    # biblecontent_id = kwargs.get("bible_content_id",None)
+    bibleaudio_id = kwargs.get("bibleaudio_id",None)
     if source_name not in db_models.dynamicTables:
         raise NotAvailableException(f'{source_name} not found in database.')
     if not source_name.endswith('_bible'):
         raise TypeException('The operation is supported only on bible')
     model_cls = db_models.dynamicTables[source_name]
-    # model_cls_audio = db_models.dynamicTables[source_name+"_audio"]
+    model_cls_audio = db_models.dynamicTables[source_name+"_audio"]
     query = db_.query(model_cls).options(joinedload(model_cls.book))
     fetched = None
     if biblecontent_id:
         query = query.filter(model_cls.bookContentId  == biblecontent_id)
+    if bibleaudio_id:
+        query = query.filter(model_cls_audio.audioId  == bibleaudio_id)
     if book_code:
         query = query.filter(model_cls.book.has(bookCode=book_code.lower()))
     if content_type == "usfm":
--- a/app/crud/structurals_crud.py
+++ b/app/crud/structurals_crud.py
@@ -144,30 +144,41 @@ def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
         "licenses":db_models.License,
         "versions":db_models.Version,
         "content_types": db_models.ContentType,
-        "sources":db_models.Source,
-        "infographic":db_models.Infographic}
+        "sources":db_models.Source}
     if db_restore.deletedFrom in content_class_map:
         model_cls = content_class_map[db_restore.deletedFrom]
     else:
-        if not  get_sources(db_, table_name=db_restore.deletedFrom):
-            raise NotAvailableException('Source not found in database')
-        source = get_sources(db_, table_name=db_restore.deletedFrom)[0]
-        source_name = source.sourceName
-        if source_name.endswith("bible") is False:
-            model_cls = db_models.dynamicTables[source.sourceName]
+        if db_restore.deletedFrom.endswith("audio"):
+            renamed_table = db_restore.deletedFrom.replace("_audio", "")
+            if not  get_sources(db_, table_name = renamed_table):
+                raise NotAvailableException('Source not found in database')
+        elif db_restore.deletedFrom.endswith("cleaned"):
+            renamed_table = db_restore.deletedFrom.replace("_cleaned", "")
+            if not  get_sources(db_, table_name = renamed_table):
+                raise NotAvailableException('Source not found in database')
+        if db_restore.deletedFrom.endswith(("audio","cleaned")) is False:
+            if not  get_sources(db_, table_name=db_restore.deletedFrom):
+                raise NotAvailableException('Source not found in database')
+            source = get_sources(db_, table_name=db_restore.deletedFrom)[0]
+            source_name = source.sourceName
+            if source_name.endswith("bible") is False:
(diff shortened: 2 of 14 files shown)


app/crud/contents_crud.py
```diff
@@ -838,23 +856,24 @@ def get_bible_versification(db_, source_name):
         }
     return response
 
-def get_available_bible_books(db_, source_name,book_code=None, content_type=None,
-    biblecontent_id=None, **kwargs):
+def get_available_bible_books(db_, source_name,book_code=None, content_type=None,#pylint: disable = too-many-arguments,disable=too-many-locals
```
sijumoncy: too-many-args pylint can be avoided by access bibleaudio_id via kwargs

app/crud/contents_crud.py
```diff
@@ -838,23 +856,24 @@ def get_bible_versification(db_, source_name):
...
     if not source_name.endswith('_bible'):
         raise TypeException('The operation is supported only on bible')
     model_cls = db_models.dynamicTables[source_name]
-    # model_cls_audio = db_models.dynamicTables[source_name+"_audio"]
+    model_cls_audio = db_models.dynamicTables[source_name+"_audio"]
     query = db_.query(model_cls).options(joinedload(model_cls.book))
     fetched = None
     if biblecontent_id:
         query = query.filter(model_cls.bookContentId  == biblecontent_id)
+    if bibleaudio_id:
```
sijumoncy, line 876: This is a common fucntion for getting bible content , based on content_type,
what happend if provide content_type other than audio and bibleaudio_id. 
Does this query returns proper response ?

app/crud/structurals_crud.py
```diff
@@ -144,29 +144,33 @@ def restore_data(db_: Session, restored_item :schemas.RestoreIdentity):
...
     if db_restore.deletedFrom in content_class_map:
         model_cls = content_class_map[db_restore.deletedFrom]
     else:
         if not  get_sources(db_, table_name=db_restore.deletedFrom):
             raise NotAvailableException('Source not found in database')
-        source = get_sources(db_, table_name=db_restore.deletedFrom)[0]
-        source_name = source.sourceName
-        if source_name.endswith("bible") is False:
-            model_cls = db_models.dynamicTables[source.sourceName]
+        if db_restore.deletedFrom.endswith(("audio","cleaned")) is False:
```
sijumoncy, line 159: this means the control go to else part when the deleted from not endswith audio or cleaned. 
I saw the audio handled in else case . what about cleaned ?

app/schema/schema_content.py
```diff
@@ -27,7 +27,7 @@ class Config:
 
 class AudioBible(BaseModel):
     '''Response object of Audio Bible'''
-    # audioId: int
+    audioId: int = None
```
sijumoncy: is this really need to make optional ? don't  we get unique id in every request

app/test/test_bibles.py
```diff
@@ -1306,3 +1306,397 @@ def test_restoreitem_with_notavailable_source():
...
+    assert response.status_code == 201
+    assert len(response.json()['data']) == 6
+    assert response.json()['message'] == "Bible audios details uploaded successfully"
+    for item in response.json()['data']:
+        assert_positive_get_for_audio(item)
+    headers_auth = {"contentType": "application/json",#pylint: disable=redefined-outer-name
+                "accept": "application/json"}
+    headers_auth['Authorization'] = "Bearer"+" "+initial_test_users['VachanAdmin']['token']
+    #Check bible audio exist in dynamic table ater post
+    post_response = client.get(UNIT_URL+source_name+"/books?content_type=audio&udio_id"+str(audio_id),headers=headers_auth)
```
sijumoncy: How this test is passed ? 
content_type=audio&udio_id -  wrong param is passed here right ?

sijumoncy: is saw the response contain 6 items . The audio id is for book wise or single audio ?

app/test/test_bibles.py
```diff
@@ -1306,3 +1306,397 @@ def test_restoreitem_with_notavailable_source():
...
+                    'Authorization': "Bearer"+" "+initial_test_users['VachanAdmin']['token']
+            }
+    response = client.delete(UNIT_URL+source_name+'/audios', headers=headers_va, json=data)
+    assert response.status_code == 200
+    assert response.json()['message'] ==\
+         f"Bible Audio with id {audio_id} deleted successfully"
+    delete_response = client.get(UNIT_URL+source_name+"/books?content_type=audio&audio_id"+str(audio_id),headers=headers_auth)
+    assert delete_response.status_code == 200
+    assert len(delete_response.json()) == 5
+    assert not delete_response.json()[0]['audio']['audioId'] == audio_id
```
sijumoncy: do you think , this need to test in a loop because 5 items are in response ?

app/test/test_bibles.py
```diff
@@ -1306,3 +1306,397 @@ def test_restoreitem_with_notavailable_source():
...
+    assert response.json()['message'] == "Login Succesfull"
+    test_user_token = response.json()["token"]
+    headers_sa= {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token
+            }
+
+    data = {"sourceName":source_name}
+    response = client.delete(UNIT_URL, headers=headers_sa, json=data)
+    assert response.status_code == 404
```
sijumoncy: this need to be call an input validation error . 1 input item is missing ?

app/test/test_bibles.py
```diff
@@ -1306,3 +1306,397 @@ def test_restoreitem_with_notavailable_source():
...
+    response = login(as_data)
+    assert response.json()['message'] == "Login Succesfull"
+    test_user_token = response.json()["token"]
+    headers_sa= {"contentType": "application/json",
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token
+            }
+    data = {"itemId":audio_id}
+    response = client.delete(UNIT_URL, headers=headers_sa, json=data)
+    assert response.status_code == 404
```
sijumoncy: this also to be missing error ?

app/test/test_bibles.py
```diff
@@ -1306,3 +1306,397 @@ def test_restoreitem_with_notavailable_source():
...
+    headers_auth = {"contentType": "application/json",#pylint: disable=redefined-outer-name
+                    "accept": "application/json",
+                    'Authorization': "Bearer"+" "+test_user_token
+            }
+    response = client.put(RESTORE_URL, headers=headers_auth, json=data)
+    assert response.status_code == 201
+    assert response.json()['message'] == \
+    f"Deleted Item with identity {deleteditem_id} restored successfully"
+    #Check bible book exists in dynamic table after restore
+    restore_response =  client.get(UNIT_URL+source_name+'/books?book_code=mat',\
```
sijumoncy: what is the content type here ? audio or all

app/test/test_bibles.py
```diff
@@ -1306,3 +1306,397 @@ def test_restoreitem_with_notavailable_source():
...
+        headers=headers_auth)
+    assert restore_response.status_code == 200
+    assert len(restore_response.json()) == 1
+    for item in restore_response.json():
+        assert_positive_get_for_books(item)
+    #check item exists in cleaned table
```
sijumoncy: delete and restore audio does affect bible and cleaned ?

app/test/test_bibles.py
```diff
@@ -1306,3 +1306,397 @@ def test_restoreitem_with_notavailable_source():
...
+            }
+    #Delete Associated Source
+    get_source_response = client.get(SOURCE_URL + "?source_name="+source_name, headers=headers_auth)
+    source_id = get_source_response.json()[0]["sourceId"]
+    data = {"itemId":source_id}
+    response = client.delete(SOURCE_URL, headers=headers_auth, json=data)
+    assert response.status_code == 200
+    #Restoring data
+    #Restore content with Super Admin after deleting source
+    restore_response = client.put(RESTORE_URL, headers=headers_auth, json=data)
```
sijumoncy: is this the same data you are passing here as the above one (delete source , restore audio )

sijumoncy (changes requested):

app/crud/structurals_crud.py
shimilgithub: > this means the control go to else part when the deleted from not endswith audio or cleaned. I saw the audio handled in else case . what about cleaned ?

On restore, cleaned table  has to be deleted along with bible books only.It should not be removed alone.Thats why that check is kept

app/test/test_bibles.py
shimilgithub: No.bible and cleaned has the info about bible books.That test should not be in bile_audios.Removed and resolved

app/schema/schema_content.py
shimilgithub: not necessary.removed "None" and resolved

sijumoncy: check does that make any issue in get bible books

app/test/test_bibles.py
shimilgithub: Resolved

app/test/test_bibles.py
shimilgithub: Resolved

app/test/test_bibles.py
shimilgithub: Did you mean same as "test_restoreitem_with_notavailable_source" ?

sijumoncy: 1695 and 1699

shimilgithub: > 

yes

shimilgithub: It should be different.resolved.

app/test/test_bibles.py
shimilgithub: changed to audio

app/test/test_bibles.py
shimilgithub: modified to fetch single item.both issues resolved

sijumoncy (approved):

Issue #502, opened 2023-02-09
changes / new APIs required for Glossary


LijuJacob08:
Issues-
-we can't list the glossary words with current [api](https://stagingapi.vachanengine.org/docs#/Nlp/get_glossary_v2_nlp_gloss_get) (as the token is a required field).

-there is no PUT call as well.(for editing the Glossary)

LijuJacob08:
Hi Kavitha ,
can we that total row count value in our response[for the new [api](https://stagingapi.vachanengine.org/docs#/Nlp/get_glossary_entries_v2_nlp_gloss_entries_get)]..? this will be helpful for me to implement serverside pagination.
thanks

LijuJacob08:
Adding one more to the list:-
can we have token list only query for the glossary something similary to word_list_only available for dictionary.

have one question ->do we have active and inactive tokens here as well like we have it for dictionary words?

kavitharaju:
- [x] new GET list of glossary entires
- [x] new GET count of total glossary entries, with option to include/exclude inactive ones
- [x] change the POST api for learn gloss to not add metadata in a different entry but itself.
- [x] new PUT api to edit glossary items
- [x] new DELETE API for glossary items

PR #507, opened 2023-03-06
Feature/glossary list api

kavitharaju:
- Implementation of the GET API as per the request #502 
- The API lists all entries in the translation memory for a given language pair.
- Skip and Limit values are by default `null`. So all entries will be fetched. Recommends use of skip and limit
- Token, is an optional query param. Upon providing it, only those entries corresponding to it will be fetched.


kavitharaju:
@LijuJacob08 
See if this API would meet your needs. Do you have any suggestions?

sijumoncy (approved):

shimilgithub (approved):

PR #510, opened 2023-03-10
Hard delete on Autographa Project and Users

shimilgithub:
Issue #405


app/auth/access_rules.json
```diff
@@ -38,7 +38,9 @@
     },
     "translation-project":{
         "create":["SuperAdmin", "AgAdmin", "AgUser"],
+        "delete":["SuperAdmin", "AgAdmin", "AgUser"],
```
kavitharaju: We shouldn't let any AgUser delete the project. Only projectOwner should be able to delete a specific project. so we need the following
```
"delete":["SuperAdmin", "AgAdmin", "projectOwner"],
```

app/crud/structurals_crud.py
```diff
@@ -101,13 +101,17 @@ def update_language(db_: Session, lang: schemas.LanguageEdit, user_id=None):
...
-    source = None, user_details = None):
+    source = None,deleting_user=None):
     '''backup deleted items from any table'''
     json_string = jsonpickle.encode(del_content)#, unpicklable=False
     json_string=json.loads(json_string)
     del json_string['py/object'],json_string['_sa_instance_state']
+    try:
+        del_item_createduser = del_content.createdUser
+    except Exception: # pylint: disable=W0703
+        del_item_createduser = deleting_user
```
kavitharaju: I made this change, if I remember correctly. But I did that to temporarily get the code working and to not affect your older APIs. 
We dont want 
```
del_item_createduser = del_content.createdUser
```
This is incorrect info. When we record a transaction, the person who did that action should be recorded. In this case the user who is calling the deleteAPI shoudl be recorded. Not the value from createdUser field.

app/db_models.py
```diff
@@ -402,8 +402,8 @@ def project_id(cls): # pylint: disable=E0213
     @declared_attr
     def project(cls): # pylint: disable=E0213
         '''For modelling the targetLanguage field in this class'''
-        return relationship(TranslationProject, foreign_keys=cls.project_id, uselist=False,
-            back_populates="users")
+        return relationship(TranslationProject, foreign_keys=cls.project_id, uselist=False)
+            # back_populates="users")
```
kavitharaju: What is the need of this change? I am just asking to understand.

app/routers/content_apis.py
```diff
@@ -74,32 +74,10 @@ async def delete_contents(request: Request, content_obj: schemas.DeleteIdentity
         raise NotAvailableException(f"Content id {content_id} not found")
     deleted_content = structurals_crud.delete_content(db_=db_, content=content_obj)
     delcont = structurals_crud.add_deleted_data(db_=db_,del_content= deleted_content,
-            table_name = dbtable_name,user_details=user_details)
+            table_name = dbtable_name)
```
kavitharaju: Dont we need to pass the details of current user to be recorded in deleted_items table?

kavitharaju: I see same action done on several cases below this. I think passing those `user_details` are needed.

kavitharaju: I think it is to be passed as `deleting_user=user_details['user_id']`

kavitharaju (changes requested):

app/db_models.py
shimilgithub: > What is the need of this change? I am just asking to understand.

This change was there in referred code.

shimilgithub: reverted changes

kavitharaju: Is everything working properly after the revert?
I dont remember clearly, but I doubt I might have made this change to fix some issues. Are all tests running correctly on local too?

shimilgithub: > Is everything working properly after the revert? I dont remember clearly, but I doubt I might have made this change to fix some issues. Are all tests running correctly on local too?

Yes.Everything is working fine

kavitharaju: alright then

kavitharaju (approved):

PR #512, opened 2023-03-15
Hard delete on sentences

shimilgithub:
Issue #405

### DIFF
--- a/app/auth/access_rules.json
+++ b/app/auth/access_rules.json
@@ -46,7 +46,7 @@
         "edit-draft": ["SuperAdmin", "AgAdmin", "projectOwner", "projectMember"],
         "read-draft":["SuperAdmin", "AgAdmin", "projectOwner", "projectMember", "BcsDeveloper"],
         "view-project":["SuperAdmin", "AgAdmin", "projectOwner", "projectMember", "BcsDeveloper"],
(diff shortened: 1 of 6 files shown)


app/auth/access_rules.json
```diff
@@ -45,7 +45,8 @@
         "read-settings":["SuperAdmin", "AgAdmin", "projectOwner", "projectMember"],
         "edit-draft": ["SuperAdmin", "AgAdmin", "projectOwner", "projectMember"],
         "read-draft":["SuperAdmin", "AgAdmin", "projectOwner", "projectMember", "BcsDeveloper"],
-        "view-project":["SuperAdmin", "AgAdmin", "projectOwner", "projectMember", "BcsDeveloper"]
+        "view-project":["SuperAdmin", "AgAdmin", "projectOwner", "projectMember", "BcsDeveloper"],
+        "delete-sentence": ["SuperAdmin", "AgAdmin", "projectOwner"]
```
kavitharaju: We need to let all users who can edit draft also to delete-sentences. So use the same list of roles here.

kavitharaju (changes requested):
Looks good to me. Just one small change in access-rule

kavitharaju (approved):

PR #516, opened 2023-03-17
Hard delete on glossary

shimilgithub:
Issue #405

### DIFF
--- a/app/crud/nlp_crud.py
+++ b/app/crud/nlp_crud.py
@@ -820,6 +820,32 @@ def agmt_suggest_translations(db_:Session, project_id, books, sentence_id_range,
         }
     return response
 
+def remove_glossary(db_, source_lang,target_lang, token):
+    '''To remove a suggestion'''
+    if isinstance(source_lang, str):
+        source_lang = db_.query(db_models.Language).filter(
+            db_models.Language.code == source_lang).first()
+        if not source_lang:
+            raise NotAvailableException("Source language not available")
+    if isinstance(target_lang, str):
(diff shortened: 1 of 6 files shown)


app/crud/nlp_crud.py
```diff
@@ -820,6 +820,32 @@ def agmt_suggest_translations(db_:Session, project_id, books, sentence_id_range,
...
+    if isinstance(target_lang, str):
+        target_lang = db_.query(db_models.Language).filter(
+            db_models.Language.code == target_lang).first()
+        if not target_lang:
+            raise NotAvailableException("Target language not available")
+
+    token_row = db_.query(db_models.TranslationMemory).filter(
+            db_models.TranslationMemory.source_lang_id == source_lang.languageId,
+            db_models.TranslationMemory.target_lang_id == target_lang.languageId,
+            db_models.TranslationMemory.token == token).first()
```
kavitharaju: I doubt if matching with just the token and taking the first as the row to be deleted is a correct assumption. Because there might be multiple entries for the same token. So what identifies a unique row needs to be decided.

I dont think even a token-translation pair will be able to uniquely identify the row to be deleted. We may have to go by ID field. But its worth looking into the tables insertion logic before we decide that.

Can you check the logic of 
 - POST glossary API, 
 - learn alignment API and 
 - the two APIs for applying token translation in project and generic. 
  
 I think these are the places from where this table gets updated. Try to understand how all it may get inserted to and updated. Then we can decide how to let users delete.

kavitharaju (changes requested):

kavitharaju (approved):

Issue #457, opened 2022-12-02
Refactor Authentication and Implement New Apis


sijumoncy:
2. ### Basic Refactor
    - [x] Authentication Change to DB Based from File Based
    - [x] update Seed Db scripts for default values and tables
    - [x] Moving permission map, access rules and corresponding data to DB
    - [x] Permission map, access rules and other required, default values added in csv to upload through seed script
    - [x]  Implemented Functions to load required data from DB
    - [x] Refactored authentication module for the changes
    - [x] Necessary changes in test config and test cases
    - [x] Endpoints for App ( Register, Generate Key, Remove key, Delete App, Edit )
    - [x] Removed Header based app identification to key through query params. Query params for key added for all endpoints
    - [x] kratos instance and schema created for handling app (only working on local, need to fix other ymls) ( using V10.1 )
    - [x] Remove API-App from default user and make email and password from ENV
    - [x] Fix docker yml is not running - can not connect to second kratos instance (test , docker-compose)
    - [x] Filter App schema remove ( used in get users api for filter list ) ,Based on roles from db-kratos
    - [x] Change AdminRole schema enum to db based logic
    - [x] App and User toke only allow to use in respective endpoints for login, logout etc.
    - [x] update role db with app name (none field), description and update code according to it
- - - -
2. ### API Implementation
    - [x] Roles APIs
                 - POST, PUT, GET, DELETE (if a role is deleted it wont create an issue on new role create with same name, but create 
                   issue on restore )
                 - App name is a must param in create, put role and not allow to change app name later
                 - Implement Test cases for Roles
    - [ ] App APIS
                 - APIs are already created , need to pass default role when register a new app
                 - [ ] create / update app details on db table
                 - Do not allow to edit default role field in edit ( ensure the implementation Done )
                 - Implement test cases for apps
                 - test cases for login, logout, delete functionality apis of user and app with their tokens and ensure the app token not 
                    work in user endpoints and vice versa.
    - [x] Permission APIs
                 - POST , PUT, GET, DELETE ( Deleted restore may cause issue for unique name)
                 - Add appName field to permission table in db to identify the permission belongs to which app, description also if 
                    needed
                 - Implement Test cases
    - [x] Access Rules
                 - POST, PUT, GET, DELETE
                 - Implement Test cases
    - [x] Permission Map
                 - POST, PUT, GET, DELETE
                 - Implement Test cases
[...]

PR #518, opened 2023-03-24
Access Rules Post

sijumoncy:
- Post Api implementation for Access Rules
- #457


kavitharaju:
@sijumoncy Could you please remind me, did we make a decision about gql_tests? I see them failing...

app/crud/auth_crud.py
```diff
@@ -100,3 +101,34 @@ def update_role(db_: Session, role_details, user_id=None):
...
+        db_role = db_.query(db_models.Roles).filter(
+            func.lower(db_models.Roles.roleName) == func.lower(role.strip())).first()
+        if not db_role:
+            raise NotAvailableException(f"role, {role.strip()},"+\
+                " not found in Database")
+
+    db_content = db_models.AccessRules(
+        entitlementId=entitlement.resourceTypeId,
+        tagId=tag.permissionId,
+        roles=details.roles,
```
kavitharaju: Is roles a multi-valued field? Did we make that so in the DB schema? Shouldn't we make them multiple rows, I wonder?

app/routers/auth_api.py
```diff
@@ -375,7 +375,29 @@ async def update_roles(role_details:schema_auth.RoleIn,request: Request,#pylint:
...
+user_details =Depends(get_user_or_none),
+db_: Session = Depends(get_db)):#pylint: disable=unused-argument
+    '''Create Authentication Access Rules
+    * entitlement (ResourceType) is mandatory
+    * tag (permission) is mandatory
+    * roles (List of role) is mandatory'''
+    log.info('In create Access Rules')
+    log.debug('Access rule  Create In:%s',details)
+    data = create_access_rules(db_, details, user_id = user_details['user_id'])
+    print("returned data ------------->", data)
```
kavitharaju: could avoid the prints.

sijumoncy:
> @sijumoncy Could you please remind me, did we make a decision about gql_tests? I see them failing...

Yes , We decided to avoid those temporally, using the command

kavitharaju:
Okay then please change the test running command in the CI/CD.

app/crud/auth_crud.py
sijumoncy: We are keeping the roles as an array for a pair of entitlement and tagId. 
If needed we can change it to multiple line  like 
E1, T1, R1 - E1 , T1, R2 (where E1, T1, R1 will be a constrain unique )

kavitharaju: Yes, that is how we usually do it in DBs.

kavitharaju (approved):

kavitharaju:
@sijumoncy I see the attempt to use --ignore-glob in the pytest command hasn't worked. Another one you can try is this
`pytest -k "not test_gql"`.

But merging this PR for now.

PR #523, opened 2023-03-29
Periodic clean-up of database

shimilgithub:
Issue #405

### DIFF
--- a/app/crud/structurals_crud.py
+++ b/app/crud/structurals_crud.py
@@ -521,3 +521,29 @@ def get_bible_books(db_:Session, book_id=None, book_code=None, book_name=None,
     if book_name is not None:
         query = query.filter(db_models.BibleBook.bookName == book_name.lower())
     return query.offset(skip).limit(limit).all()
+
+def get_deleted_items(db_: Session, item_id: int =None,
+    skip: int = 0, limit: int = 100):
+    '''Fetches deleted items from deleted_items table'''
+    query = db_.query(db_models.DeletedItem)
+    if item_id is not None:
+        query = query.filter(db_models.DeletedItem.itemId == item_id)
+    return query.offset(skip).limit(limit).all()
+
+def delete_deleteditems(db_: Session):
+    '''Cleanup entire deleted_items table  and remove irrelevant dangling dynamic tables'''
+    for table_name in engine.table_names():
+        if table_name.startswith('table'):
+            if table_name.endswith("audio"):
+                tb_name = table_name.replace("_audio", "")
+            elif table_name.endswith("cleaned"):
+                tb_name = table_name.replace("_cleaned", "")
+            else:
+                tb_name = table_name
+            if not  get_sources(db_, table_name = tb_name):
+                with engine.connect() as conn:
+                    conn.execute(f"DROP TABLE {table_name} CASCADE;")
+                # table = Table(table_name, metadata, autoload=True)
+                # table.drop(engine, checkfirst=True)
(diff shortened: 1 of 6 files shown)


app/crud/structurals_crud.py
```diff
@@ -521,3 +521,29 @@ def get_bible_books(db_:Session, book_id=None, book_code=None, book_name=None,
...
+    return query.offset(skip).limit(limit).all()
+
+def delete_deleteditems(db_: Session):
+    '''Cleanup entire deleted_items table  and remove irrelevant dangling dynamic tables'''
+    for table_name in engine.table_names():
+        if table_name.startswith('table'):
+            if table_name.endswith("audio"):
+                tb_name = table_name.replace("_audio", "")
+            elif table_name.endswith("cleaned"):
+                tb_name = table_name.replace("_cleaned", "")
```
kavitharaju, line 540: I see tables named like `table_1001`, `table_1001_auido` and `table_1001_cleaned` being created in DB. I this logic where you are replacing the trailing part of name and using delete Cascade, I see the base table name will be used and all three tables will get deleted in one go. 

An issue I see with this is, even though you delete all 3 tables upon seeing the first table name in the set, the raimining two tables are also going to come. And at that time, the delete command will be run with that table not being there. Wouldn't this cause problem? Please explian, if there is anything that I am missing in this logic.

kavitharaju: But if that indeed the case couple of alternate methods you can try
1. If you see, its an audio table or a cleaned bible table, dont proceed with the rest of the steps in the for loop
2. Or, you could use "DROP TABLE IF EXISTS" command.

kavitharaju (changes requested):
I am not sure if this is the issue that is causing your tests to fail. But worth giving it a try

app/crud/structurals_crud.py
shimilgithub: > I see tables named like `table_1001`, `table_1001_auido` and `table_1001_cleaned` being created in DB. I this logic where you are replacing the trailing part of name and using delete Cascade, I see the base table name will be used and all three tables will get deleted in one go.
> 
> An issue I see with this is, even though you delete all 3 tables upon seeing the first table name in the set, the raimining two tables are also going to come. And at that time, the delete command will be run with that table not being there. Wouldn't this cause problem? Please explian, if there is anything that I am missing in this logic.

Those three tables are not deleting together.They are treated as separate table on delete.
 eg:` table_1001_audio `and `table_1001_cleaned` are renamed to `table_1001` just to check the existence of source.These renaming is assigned to different variable(`tb_name`).But deletion is performed on the original table name (`table_name`)

kavitharaju: Then why do we need CASCADE?

shimilgithub: > Then why do we need CASCADE?

When I list dynamic tables,its not coming in any particular order.If  `table_1001` comes before its audio and cleaned tables,deletion will return a database error as follows due to foreign key dependency.
" `cannot drop table table_1001 because other objects depend on it\n  constraint table_1001_audio_book_id_fkey on table table_1001_audio depends on table table_1001` "

kavitharaju: What CASCADE delete do is, it deletes not only that table but also all the tables dependent on it.

shimilgithub: > What CASCADE delete do is, it deletes not only that table but also all the tables dependent on it.

Oh.OK.Understood.I'll modify

PR #530, opened 2023-04-03
GET count of glossary entries

shimilgithub:
Issue #502


app/auth/api-permissions.csv
```diff
@@ -122,6 +122,7 @@ Endpoints,Method,RequestApp,FilterResults,ResourceType,Permission
 /v2/translation/draft,PUT,None,False,None,process
 /v2/nlp/gloss,GET,None,False,None,read
 /v2/nlp/gloss-entries,GET,None,False,None,read
+/v2/nlp/gloss-count,GET,None,False,None,read
```
kavitharaju: Can we have the end point `/v2/nlp/gloss-entries/count`? To make some common pattern with dictionary APIs.

app/routers/translation_apis.py
```diff
@@ -500,8 +500,35 @@ async def get_glossary_entries(request: Request,
...
+@get_auth_access_check_decorator
+async def get_gloss_count(request: Request,
+    source_language:schemas.LangCodePattern=Query(...,example="en"),
+    target_language:schemas.LangCodePattern=Query(...,example="hi"),
+    token:str=Query(None,example="love"),
+    user_details =Depends(get_user_or_none), db_:Session=Depends(get_db)):
+    '''Counts all glossary entries in translation memory between two languages.
+        * Also counts unique tokens in translation memory
+   `    * Can filter with or without search word/token
+        * Source anf target language should be specified
```
kavitharaju: Typo "anf" instead of "and".

app/routers/translation_apis.py
```diff
@@ -500,8 +500,35 @@ async def get_glossary_entries(request: Request,
...
+        * "gloss_entries" in response gives the count of total glossary records
+        * "token_count" in response gives the count of unique tokens'''
```
kavitharaju: I think we can may have to explain it so that a user can understand.

"gloss entries" in response counts different translations of same word as different.
"token_count" in response doesn't consider the multiple translations but give the count of unique source tokens.

app/test/test_nlp_glossary.py
```diff
@@ -69,3 +70,59 @@ def test_get_gloss_entries():
...
+        assert response.status_code == 200
+        assert isinstance(response.json(), dict)
+        assert len(response.json()) == 2
+        assert "gloss_entries" in response.json()
+        assert "token_count" in response.json()
+
+    # With filter for token - Positive Test
+    resp = client.get(GET_URL+"&token=love", headers=headers_auth)
+    assert resp.status_code == 200
+    assert resp.json()['gloss_entries'] == 2
```
kavitharaju: Can we also have tests to make sure the token_count logic also works correctly. assert it is equal to one always when input toke is specified.

kavitharaju: Also add additional checks for gloss_entries count and token_count when input doesn't have token.

app/routers/translation_apis.py
```diff
@@ -500,8 +500,35 @@ async def get_glossary_entries(request: Request,
     log.info('In get_glossary_entries')
     log.debug('source_language:%s, target_language:%s, token:%s',
         source_language, target_language, token)
-    return nlp_crud.get_glossary_list(db_, source_language, target_language, token,
+    response = nlp_crud.get_glossary_list(db_, source_language, target_language, token,
     skip=skip, limit=limit)
+    return response['gloss_entries']
+
+@router.get('/v2/nlp/gloss-count',
+    responses={502: {"model": schemas.ErrorResponse},
```
kavitharaju, line 510: Dont we have response model defined  for this? We should. that will create documentation in swagger automatially and also perform the validations by default. 

And I think we can call the two fields `token_translation_count` and `token_count` which will be more self explanatory that we do consider translation in one and do not consider it in the other.

kavitharaju (changes requested):

app/routers/translation_apis.py
```diff
@@ -502,9 +502,10 @@ async def get_glossary_entries(request: Request,
         source_language, target_language, token)
     response = nlp_crud.get_glossary_list(db_, source_language, target_language, token,
     skip=skip, limit=limit)
-    return response['gloss_entries']
+    return response['token_translation_count']
 
-@router.get('/v2/nlp/gloss-count',
+@router.get('/v2/nlp/gloss-entries/count',
+    response_model= dict,
```
kavitharaju: Just dict is not enough. We need to have a Pydantic model defined in the schemas/ and field names and datatypes specified. then only the API dcoumentaion will be of use to the users. Also, then only the validation will be more meaningful

kavitharaju (changes requested):

kavitharaju (approved):

Issue #532, opened 2023-04-04
Document error in language code format


shimilgithub:
In the API documentation of POST Language (swagger and redoc) we need to specify the acceptable format of language code to BCP-47.

Issue #531, opened 2023-04-04
Incorrect tag for DELETE APIs


shimilgithub:
Change "`Autographa-Project management`" and "`Autographa-Translation`" tags of DELETE APIs to "`Translation-Project management`" and "`Project-Based-Translation`" respectively.

PR #534, opened 2023-04-05
Change the POST API for learn gloss by adding metadata in same entry

shimilgithub:
Issue #502 

Commit also include:
Issue #531 
Issue #532

### DIFF
--- a/app/crud/nlp_crud.py
+++ b/app/crud/nlp_crud.py
@@ -481,20 +481,24 @@ def rebuild_trie(db_, src, trg):
             json_data = json.load(json_file)
             for key in json_data:
                 new_trie[key] = json_data[key]
-    # display_tree(new_trie)
     return new_trie
 
-def add_metadata_to_first_occurance(db_, gloss, source_lang, target_lang, *args):
-    """Always add meta data to the first occurance of a source token only'''"""
+def add_metadata(db_, gloss, source_lang, target_lang, *args):
+    """Always add meta data to the respective token-translation pair"""
     recieved_args = args[0]
     db_content = args[1]
     if 'tokenMetaData' in gloss and gloss['tokenMetaData'] is not None:
-        # Always add meta data to the first occurance of a source token only'''
-        token_row = db_.query(db_models.TranslationMemory).filter(
+        query_filter = db_.query(db_models.TranslationMemory).filter(
             db_models.TranslationMemory.source_lang_id == source_lang.languageId,
             db_models.TranslationMemory.target_lang_id == target_lang.languageId,
             db_models.TranslationMemory.token == gloss['token']).order_by(
-            db_models.TranslationMemory.tokenId.asc()).first()
+            db_models.TranslationMemory.tokenId.asc())
+        if 'translations'in gloss['tokenMetaData']:
+            token_row = query_filter.filter(
+                db_models.TranslationMemory.translation ==
+                    gloss['tokenMetaData']['translations']).first()
+        else:
+            token_row = query_filter.first()
         if token_row:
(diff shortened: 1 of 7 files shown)


app/crud/nlp_crud.py
```diff
@@ -481,20 +481,24 @@ def rebuild_trie(db_, src, trg):
             json_data = json.load(json_file)
             for key in json_data:
                 new_trie[key] = json_data[key]
-    # display_tree(new_trie)
     return new_trie
 
-def add_metadata_to_first_occurance(db_, gloss, source_lang, target_lang, *args):
-    """Always add meta data to the first occurance of a source token only'''"""
+def add_metadata(db_, gloss, source_lang, target_lang, *args):
+    """Always add meta data to the respective token-translation pair"""
```
kavitharaju: Now that we are doing it this way, do we really need a separate function for handling metadata? Can't we just use the calling function itself for this?

app/schema/schemas_nlp.py
```diff
@@ -220,14 +220,14 @@ class GlossInput(BaseModel):
     '''Import object for glossary(dictionary) data for learning'''
     token: str = Field(..., example="love")
     translations: List[str] = Field(None, example=['प्यार', "प्रेम", "प्रेम करना"])
-    tokenMetaData: dict = Field(None, example={"word-class":["noun", "verb"]})
+    tokenMetaData: dict =Field(None, example={"translations":"प्रेम","word-class":["noun", "verb"]})
```
kavitharaju: Dont we have a field named translations? What is the reason for adding it as an example in metadata?

app/test/test_translation_suggestions.py
```diff
@@ -306,6 +304,46 @@ def test_bug_fix():
     assert list(response.json()["translations"].keys())[0] == "Abraham's".lower()
     assert response.json()["metaData"]["for"] == "अब्राहम की"
 
+def test_add_metadata_to_same_gloss():
+    '''testing metadata is added to the same token-translation pair'''
+    tokens_test = [
+        {"token":"test", "translations":["ടെസ്റ്റ്"],"tokenMetaData":{"translations":"ടെസ്റ്റ്","tense": "present"}},
+        {"token":"tested", "translations":["ടെസ്റ്റഡ്", "ടെസ്റ്റ് ചെയ്തു"],"tokenMetaData":{"translations":"ടെസ്റ്റ് ചെയ്തു","tense": "past"}}
```
kavitharaju: Oh... now I notice that the POST is written to accept one-source-token and multiple translations in one object. This logic will not work with the current change. I see that you are expecting the user to give which translation, in metadata, but that will not allow them to add different metadata to different translations.
We need to make this POST to add just one translation at a time. 
```
[
{ token: "test", translation: "ടെസ്റ്റ്", metaData:{...}},
{ token: "test", translation: "ടെസ്റ്റ്2", metaData:{...}},
...
]
```

This may impact not only the API, but also internal places where this would be used like in learn alignment etc

kavitharaju (changes requested):
In the GET glossary functionality also there is some extra handling done for combined meta data. We may have to remove that too
https://github.com/Bridgeconn/vachan-api/blob/81c490d17145fe25f6a4d0cd18489126f5d02968/app/crud/nlp_crud.py#L709

app/schema/schemas_nlp.py
kavitharaju: Also, its better to change the field name to "metaData" instead of "tokenMetaData" to reflect the change in its meaning with this new rewrite

kavitharaju (approved):
Looks good.

PR #537, opened 2023-04-13
PUT API for Glossary

shimilgithub:
Issue #502

### DIFF
--- a/app/crud/nlp_crud.py
+++ b/app/crud/nlp_crud.py
@@ -510,9 +510,7 @@ def add_translation_memory_gloss_dataprocess(db_, gloss_list, source_lang, #pyli
                     db_models.TranslationMemory.translation == trans).first()
                 if token_row:
                     token_row.frequency += freq_val
-                    if not token_row.metaData:
-                        token_row.metaData = {}
-                    else:
+                    if token_row.metaData:
                         for key in gloss['metaData']:
                             token_row.metaData[key] = gloss['metaData'][key]
                         flag_modified(token_row, 'metaData')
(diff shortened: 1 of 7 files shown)


app/crud/nlp_crud.py
```diff
@@ -811,6 +811,21 @@ def project_suggest_translations(db_:Session, project_id, books, sentence_id_ran
...
+    db_content = query.filter(db_models.TranslationMemory.tokenId == token_info.tokenId).first()
+    if db_content is None:
+        raise NotAvailableException(f"Token with id {token_info.tokenId} is not available.")
+    if token_info.translation is not None:
+        db_content.translation = token_info.translation
+        db_content.translationRom = utils.to_eng(token_info.translation)
+    if token_info.metaData is not None:
+        db_content.metaData = token_info.metaData
+    else:
+        db_content.metaData = {}
```
kavitharaju: Does this means, metadata will be set to {} if not provided?

Scenario; User have the following content in DB, [token1, wrong_trans, correct_metadata]
Then he uses PUT API to correct just the translation by uploading [ id, new_trans]
Wouldn't the metaData become `{}` after this PUT call, which shouldn't happen?

app/schema/schemas_nlp.py
```diff
@@ -234,6 +234,7 @@ class Config:
 
 class TranslationMemoryOut(BaseModel):
     '''Output object to directly examine Translation memory entries'''
+    tokenId: int = Field(..., example= 100000)
```
kavitharaju: Lets not call it `tokenId` because token is not unique. Something like `tmID`, `translatiionMemoryId` or simply `itemId` will do.

kavitharaju (changes requested):

kavitharaju (approved):

Issue #535, opened 2023-04-12
Unwanted files in the code base


kavitharaju:
Noticed these files and folders in the repo that shouldn't have been pushed to it. This might have crept in when we had not made the version-2 branch protected and accidentally pushed/merged to it without peer review.

These are the ones noticed, but its worth taking a look at the whole repo to see if there are more
- https://github.com/Bridgeconn/vachan-api/blob/version-2/app/deleted_contents.json
- https://github.com/Bridgeconn/vachan-api/blob/version-2/package-lock.json
- https://github.com/Bridgeconn/vachan-api/tree/version-2/docker/vachan-ENV
- https://github.com/Bridgeconn/vachan-api/blob/version-2/.project

PR #544, opened 2023-04-19
Remove unwanted files from version-2 codebase

shimilgithub:
Issue #535


kavitharaju (commented):
The entire `vachan-ENV` folder should be removed? Is that done so?

shimilgithub:
> The entire `vachan-ENV` folder should be removed? Is that done so?

yes

kavitharaju (approved):

Issue #546, opened 2023-04-21
Source post, put calls labels not working 


RevantCI:
In source [post](https://api.vachanengine.org/docs#/Sources/add_source_v2_sources_post), [put](https://api.vachanengine.org/docs#/Sources/edit_source_v2_sources_put) calls, labels not working as exptected.
If we remove the labels key the below sample post call body works
Sample Post call body
```
{
  "contentType": "bible",
  "language": "hi",
  "version": "ERV",
  "versionTag": "1",
  "labels": ["latest"],
  "year": 2020,
  "license": "ISC",
  "accessPermissions": [
    "content"
  ],
  "metaData": {
    "otherName": "KJBC, King James Bible Commentaries"
  }
}
```
Sample response 
```
{
  "error": "Error",
  "details": "ARRAY.contains() not implemented for the base ARRAY type; please use the dialect-specific ARRAY type"
}
```

PR #548, opened 2023-04-26
Bug Fix of Source Label in  POST and PUT APIs

shimilgithub:
Issue #546


kavitharaju (approved):

kavitharaju:
It is a practice we follow to include tests with all the bug fixes we do. To ensure the issue has been fixed and also to future proof against the same issue repeating and remaining uncaught upon some later changes.

Issue #526, opened 2023-03-31
Delete of deleted contents


kavitharaju:
1. The contents deleted from tables are now moved to a deleted items table. This table can be cleanup periodically by super Admin to save space.
2. The test runs as well as deletion of sources, leave dangling tables as residue. These are also to be removed upon removing contents from the deleted_items table

shimilgithub:
### Testing results of database clean-up
- **Test 1** 

_Action:_ Created dynamic tables by running bible tests and tried to delete all dangling tables (does not have an existing source) through swagger.  

_Result:_ **Test Passed**. All the dangling tables generated by “test_bible” are removed. 

- **Test 2:** 

_Action:_ Created dynamic tables by running bible tests. Wrote testcase for creating and deleting a language (or any metacontent resource which does not create dynamic table) and executed this test.  

_Result:_ **Test Passed**. All the dangling tables generated by “test_bible” are removed. 

- **Test 3:** 

_Action:_  Wrote testcase for creating and deleting a commentary (This will create dynamic tables) and tried to remove those dangling tables. 

_Result:_  **Test Failed**. 

Code get freeze while running this testcase at the line of “delete operation” (conn.execute(f"DROP TABLE {table_name} CASCADE;"). 

**Reason Identified:** 

The table which is trying to delete (in test case) is in use. Hence the database is preventing from deleting the table, and the query could get stuck waiting for a response from the database. 

Checked if there is any lock on the table using  pg_stat_activity  and identified that the connection is holding a lock on the table which need to delete. 

Tried to terminate the query by running the pg_terminate_backend() function with the corresponding pid value but it closes the entire connection.

PR #560, opened 2023-05-08
periodic cleanup of database

shimilgithub:
- #526

### DIFF
--- a/app/crud/structurals_crud.py
+++ b/app/crud/structurals_crud.py
@@ -10,6 +10,7 @@ from sqlalchemy.orm import Session
 from sqlalchemy.orm.attributes import flag_modified
 from sqlalchemy.sql import text
 from sqlalchemy.dialects.postgresql import array
+from sqlalchemy import MetaData
 import db_models
 from schema import schemas
 from custom_exceptions import NotAvailableException, TypeException, AlreadyExistsException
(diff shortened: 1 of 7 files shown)


app/crud/structurals_crud.py
```diff
@@ -584,3 +585,21 @@ def get_bible_books(db_:Session, book_id=None, book_code=None, book_name=None,
...
+        if table_name.startswith('table'):
+            if table_name.endswith("audio"):
+                tb_name = table_name.replace("_audio", "")
+            elif table_name.endswith("cleaned"):
+                tb_name = table_name.replace("_cleaned", "")
+            else:
+                tb_name = table_name
+            if not  get_sources(db_, table_name = tb_name):
+                with engine.connect() as conn:
+                    conn.execute(f"DROP TABLE IF EXISTS {table_name} CASCADE;")
```
kavitharaju, line 603: Will there be any problem in the logic if we avoid `CASCADE`? Seeing that is a bit scary...

app/test/test_deleted_items.py
```diff
@@ -0,0 +1,175 @@
+'''Test cases for deleted items related APIs'''
```
kavitharaju: Please add a line in the docstring also as to why the tests are failing and why we are commenting them out for future record.

app/crud/structurals_crud.py
shimilgithub: It is required to remove dangling tables of "bible". If "bibles" table comes first (before _audio and _cleaned tables),it will get stuck without deleting the table because of dependency.

kavitharaju: Sorry, i had forgotten we had this conversation.

kavitharaju (approved):

PR #561, opened 2023-05-08
permission map api - Post

sijumoncy:
- #457 
- endpoint for permission-map POST

### DIFF
--- a/app/routers/auth_api.py
+++ b/app/routers/auth_api.py
@@ -13,7 +13,7 @@ from auth.authentication import user_register_kratos,login_kratos,user_role_add 
 from auth.auth_app import app_register_kratos, app_update_kratos, get_filter_apps
 from crud.auth_crud import create_auth_permission, update_auth_permission,\
     get_auth_permission, create_role, get_role, update_role, create_access_rules,\
-    update_access_rules, get_access_rules
+    update_access_rules, get_access_rules, create_permission_map
 
 router = APIRouter()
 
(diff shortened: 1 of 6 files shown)


app/routers/auth_api.py
```diff
@@ -13,7 +13,7 @@
 from auth.auth_app import app_register_kratos, app_update_kratos, get_filter_apps
 from crud.auth_crud import create_auth_permission, update_auth_permission,\
     get_auth_permission, create_role, get_role, update_role, create_access_rules,\
-    update_access_rules, get_access_rules
+    update_access_rules, get_access_rules, create_permission_map
```
kavitharaju: Seeing the long list of imports from the same file I wonder if using it as follows would be better
```
import crud.auth_crud as auth_crud
...
auth_crud.create_auth_permission(..)
```

app/routers/auth_api.py
```diff
@@ -448,6 +448,30 @@ async def get_auth_rules(request: Request,user_details =Depends(get_user_or_none
...
+app_key: types.SecretStr = Query(None),#pylint: disable=unused-argument
+user_details =Depends(get_user_or_none),
+db_: Session = Depends(get_db)):#pylint: disable=unused-argument
+    '''Create Authentication Permission Maps
+    * apiEndpoint - should be one of the avaialble api (mandatory)
+    * method - one from method enum (mandatory)
+    * requestApp - one of the registered app name (mandatory)
+    * resourceType - one of the available resource/entitlement name (mandatory)
+    * permission - one of the available permission/tag name (mandatory)
+    * filterResults permission map allow or not to filter the result (default False ) '''
```
kavitharaju: I think this `filterResults` would need a better explanation... You could add one more sentence to the existing sentence like this...
```
For example for an endpoint that returns a list of items like bibles in the DB, setting this to true is needed to ensure the bibles are filtered based on user's role and the bible's access permissions.
```

kavitharaju (changes requested):
2 minor suggestions. But good to clear them so that we don't forget

app/routers/auth_api.py
sijumoncy: you mean, 
Import module and call module.function ?

kavitharaju: yes, I think that is better than the long list of import we have now. But if pylint has a different opinion do as it says :)

kavitharaju: On second thoughts... may be just keep the import in multiple lines like this
```
from crud.auth_crud import create_auth_permission, update_auth_permission
from crud.auth_crud import get_auth_permission, create_role, get_role, update_role, create_access_rules
from crud.auth_crud import update_access_rules, get_access_rules, create_permission_map
```
Code looks cleaner and no need to repeat the module name all over the code.

joelthe1: To add another opinion, I usually do something like this:
```
from crud.auth_crud import (
    create_auth_permission,
    update_auth_permission,
    get_auth_permission,
    create_role, get_role, 
    update_role, 
    create_access_rules,
    update_access_rules, 
    get_access_rules, 
    create_permission_map,
)
```
```

Comment (sijumoncy) [reply app/routers/auth_api.py]: > auth_crud

I think pylint won't allow multi import from same ( use group import )

Comment (kavitharaju) [reply app/routers/auth_api.py]: Multiple import is using different libraries in same like this ` import os, re` right?

Our case is importing multiple things from same sub module. [Here](https://nickjanetakis.com/blog/comparing-3-ways-to-import-multiple-things-from-the-same-python-module) I see someone else thinking through the same options we are considering. 

I just brought it up because the present one looked like a clutter and wanted something visually 'neat'. 
```
from crud.auth_crud import create_auth_permission, update_auth_permission
from crud.auth_crud import get_auth_permission, create_role, get_role, update_role, create_access_rules
from crud.auth_crud import update_access_rules, get_access_rules, create_permission_map
```
The above method will give a _somewhat_ tidier approach without making the file too long
 ```
from crud.auth_crud import (
    create_auth_permission,
    update_auth_permission,
    get_auth_permission,
    create_role, get_role, 
    update_role, 
    create_access_rules,
    update_access_rules, 
    get_access_rules, 
    create_permission_map,
)
```
This one on the other hand keeps it much more well organized, at the cost of more lines. This has the added advantage that we can easily comment out few of the imports without much hassle while we are coding/debugging etc

kavitharaju (approved):

Issue #576, opened 2023-05-26
ISL Server Unresponsive for Suggestion API


shimilgithub:
The ISL server, is not responding after making [Suggestion API](https://islstagingapi.vachanengine.org/docs#/Translation%20Suggestion/suggest_auto_translation_v2_translation_project_suggestions_put) call. But there is no issue found in the production environment.

PR #577, opened 2023-05-29
Fix memory leak by applying join conditions inside rebuid_trie

shimilgithub:
-  #576

### DIFF
--- a/app/crud/nlp_crud.py
+++ b/app/crud/nlp_crud.py
@@ -471,9 +471,12 @@ def display_tree(tree):
 def rebuild_trie(db_, src, trg):
     '''Collect suggestions data from translation memory and traning data directory
     and rebuild the trie for language pair in memory'''
-    db_sents = db_.query(db_models.TranslationDraft, db_models.TranslationProject).filter(
-        db_models.TranslationProject.sourceLanguage.has(code = src),
-        db_models.TranslationProject.targetLanguage.has(code = trg)).all()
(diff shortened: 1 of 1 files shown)


app/crud/nlp_crud.py
```diff
@@ -471,9 +471,12 @@ def display_tree(tree):
 def rebuild_trie(db_, src, trg):
     '''Collect suggestions data from translation memory and traning data directory
     and rebuild the trie for language pair in memory'''
-    db_sents = db_.query(db_models.TranslationDraft, db_models.TranslationProject).filter(
-        db_models.TranslationProject.sourceLanguage.has(code = src),
-        db_models.TranslationProject.targetLanguage.has(code = trg)).all()
+    db_sents = db_.query(db_models.TranslationDraft, db_models.TranslationProject).\
+    join(db_models.TranslationProject, db_models.TranslationProject.projectId ==\
```
kavitharaju: Will this be a proper way to do it?
```
db_.query(db_models.TranslationDraft).join(db_models.TranslationProject).filter(...)
```
I fear putting them together in that single bracket may again lead to dot product, on top of join. 
Also, as we define foreign keys, the SQL alchemy may figure out the joining criteria on itsown.(if we are doing joins over the FK field and there is only one FK b/w those tables). So please, try if this works too

For reference: https://www.tutorialspoint.com/sqlalchemy/sqlalchemy_orm_working_with_joins.htm

kavitharaju (approved):

PR #584, opened 2023-06-02
PUT api for auth permission map

sijumoncy:
- #457 
- PUT Api for Api permission map

### DIFF
--- a/app/crud/auth_crud.py
+++ b/app/crud/auth_crud.py
@@ -264,3 +264,63 @@ def create_permission_map(db_: Session, details: schema_auth.PermissionMapCreate
         'refresh_auth_func':generate_permission_map_table
     }
     return response
+
+def update_permission_map(db_: Session, details: schema_auth.PermissionMapUpdateInput,\
+    user_id= None):#pylint: disable=too-many-branches
+    '''update a row in permission map table'''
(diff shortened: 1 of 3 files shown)


app/crud/auth_crud.py
```diff
@@ -264,3 +264,63 @@ def create_permission_map(db_: Session, details: schema_auth.PermissionMapCreate
...
+        for row in APIPERMISSIONTABLE:
+            if details.apiEndpoint.lower().strip() == row[0].lower():
+                validate_endpoint = True
+        if not validate_endpoint:
+            raise NotAvailableException(f"apiEndpoint, {details.apiEndpoint.strip()},"+\
+                " is not a valid one")
+        db_content.apiEndpoint = details.apiEndpoint.lower().strip()
+
+    if details.method:
+        db_content.method = details.method.strip()
```
kavitharaju, line 287: What is we only had end point GET xx/yy/zzz and request came for making it PUT xx/yy/zzz? Shouldn't we check and ensure that the method-path combination is valid one too?

sijumoncy: #607  for resolving this issue , endpoint, method can be taken as FK to this table by avoid editing endpoint and method here

sijumoncy:
Closing this PR . Added same PR with all changes in #639

Issue #589, opened 2023-06-12
FASTAPI  Upgradation


shimilgithub:
Currently Vachan Engine is running on an older version of `FASTAPI(0.65.2)`. It has to be upgraded to `0.95.0`. However, this upgradation presents a set of challenges that need to be addressed.

1. `TestClient.delete()` method no longer supports `payloads` in the latest FASTAPI release, thereby requiring the rewriting of all delete tests by modifying `body parameters` to `query parameters`
2.  Latest version of FastAPI is not compatible with the `graphql` module, necessitating the library migration from `graphene` to the `strawberry`.

PR #591, opened 2023-06-16
Fastapi upgradation and removing Graphql Module

AthulyaMS:
- #589 
- #605 
- #606


app/test/test_agmt_translation.py
```diff
@@ -720,21 +720,45 @@ def test_agmt_translation_access_rule_app():
...
+            "data": data_str
+        },
+        headers=headers_auth)
     assert response.status_code == 200
     assert len(response.json()) > 0
     #Without Auth from Autographa
-    response = client.get(UNIT_URL+"/token-translations?project_id="+str(project_id)+
-        "&token="+all_tokens[0]['token']+"&sentence_id=41001001&offset=0&offset=4",
-        headers=headers, json=post_obj_list)
+    data_str2 = json.dumps(post_obj_list)
```
shimilgithub: if `data_str` and `data_str` is assigned to same value, you don't need two assignment statements. Use the first one itself

app/test/test_agmt_translation.py
```diff
@@ -720,21 +720,45 @@ def test_agmt_translation_access_rule_app():
...
+        },
+        headers=headers
+        )
     assert response.status_code == 401
     assert response.json()['error'] == 'Authentication Error'
     #Without Auth and not from Autographa
-    response = client.get(UNIT_URL+"/token-translations?project_id="+str(project_id)+
-        "&token="+all_tokens[0]['token']+"&sentence_id=41001001&offset=0&offset=4",
-        json=post_obj_list)
+    data_str3= json.dumps(post_obj_list)
```
shimilgithub: This statement also can be remove

app/test/test_agmt_translation.py
```diff
@@ -1005,10 +1029,18 @@ def test_agmt_translation_access_permissions():
         assert isinstance(resp.json(), list)
         for item in resp.json():
             assert_positive_get_tokens(item)
-
-        response = client.get(UNIT_URL+"/token-translations?project_id="+str(project_id)+
-            "&token="+all_tokens[0]['token']+"&sentence_id=41001001&offset=0&offset=4",
-            headers=headers_auth, json=post_obj_list)
+        
+        data_str4= json.dumps(post_obj_list)
```
shimilgithub: here also

app/test/test_agmt_translation.py
```diff
@@ -720,21 +720,45 @@ def test_agmt_translation_access_rule_app():
...
+    ########################################
+    data_str = json.dumps(post_obj_list)
+    response = client.get(UNIT_URL + "/token-translations",
+        params={
+            "project_id": str(project_id),
+            "token": all_tokens[0]['token'],
+            "sentence_id": "41001001",
+            "offset": ["0", "4"],
+            "data": data_str
+        },
```
shimilgithub: you can reuse same code segment .no need to repeat

shimilgithub (changes requested):
Few code segments are repeated in many places. Try to reuse the same code

app/test/test_agmt_translation.py
shimilgithub: Still  `data_str` and `params` are redefined multiple times. You don't need to do that if  it contains same information. Define only once and call the same variable wherever it is required.

shimilgithub (changes requested):
Please incorporate suggestions not only in the commented place but also consistently across the code wherever you encounter the same code pattern.

Also can see follwing warning is `test bibles`.
```
/app/app/crud/contents_crud.py:897: SAWarning:SELECT statement has a cartesian product between FROM element(s) "table_xxx_audio" and FROM element "table_xxx".  Apply join condition(s) between each element to resolve.
    fetched = query.filter(model_cls.active == active).offset(skip).limit(limit).all()
```
Remove that too  by using `join operation` to avoid memory leaks in future.

shimilgithub (dismissed):

app/test/test_agmt_projects.py
```diff
@@ -1158,7 +1158,7 @@ def test_delete_user():
     # as project owner - Positive test
     user_id = initial_test_users['AgUser2']['test_user_id']
     headers_auth['Authorization'] = "Bearer"+" "+initial_test_users['AgAdmin']['token']
-    resp = client.delete(USER_URL+'?project_id='+str(new_project['projectId'])+
+    resp =client.request("delete" ,USER_URL+'?project_id='+str(new_project['projectId'])+
         '&user_id='+str(user_id),headers=headers_auth)
```
kavitharaju: I see this change consistently in tests... Can you explain the context here?
I dont see any change corresponding to this in router functions, also I see `client.get()` is still used in tests, so what is the need of changing these tests?

app/test/test_agmt_translation.py
```diff
@@ -720,21 +720,35 @@ def test_agmt_translation_access_rule_app():
...
+            "sentence_id": "41001001",
+            "offset": ["0", "4"],
+            "data": data_str
+        }
+
+    response = client.get(UNIT_URL + "/token-translations",
+        params=params,
+        headers=headers_auth)
     assert response.status_code == 200
     assert len(response.json()) > 0
```
kavitharaju, line 738: Here also can you explain what is happening?
What is the differrence between the following 
 - params=...
 - json=...
 - params={'data':...}
 Which is for query params and which for body params? I see all these being used over different tests? If all these are supported what is the need for the change?

kavitharaju (changes requested):
I see the `graphql_api` folder is still maintained even though the contents are removed. You can remove the folder too

AthulyaMS:
> I see the `graphql_api` folder is still maintained even though the contents are removed. You can remove the folder too

When folders completely deleted,there was showing some linting error in the module ,while running in repo .

app/test/test_agmt_projects.py
AthulyaMS: While ruuning tests ,there was showing error " TestClient.delete() got an unexpected keyword argument 'json'. In a documentation ,i had seen ,If we want to send a body with a DELETE method, we can replace client.delete(...) by client.request(...) .By using this ,the test passed.
202516125-8bd3c9b1-bb98-44a2-9415-6a837badc15f

app/test/test_agmt_translation.py
AthulyaMS: Here also ,while running ,there was error like " TestClient.get() got an unexpected keyword argument 'json'.  The data_str variable is created by serializing a list of Python objects (post_obj_list) into a JSON-formatted string using the json.dumps() function. This JSON string is then included as a parameter in an HTTP GET request. By modifying like this ,test passed

kavitharaju: Looking at the GET `/token-translations` API, I see it doesn't need a body parameter passed at all. It must have been a mistake in that test. I guess it can just be removed.

kavitharaju: Please check if there are other GET call tests where body params are passed unnecessarily and fix them if you can.

kavitharaju:
> When folders completely deleted,there was showing some linting error in the module ,while running in repo .

There might be linter run on that module. You can change the command for running the linter excluding that folder. And update it on `.githubworkflows/...` CI/CD pipeline to avoid that error on gitactions.

app/test/test_agmt_projects.py
kavitharaju: This is a work around for passing a `json` to delete. We would rather have the recommended standard of not using the body params in delete, than this work-around.

AthulyaMS:
#605

kavitharaju (approved):

kavitharaju:
> > When folders completely deleted,there was showing some linting error in the module ,while running in repo .
> 
> There might be linter run on that module. You can change the command for running the linter excluding that folder. And update it on `.githubworkflows/...` CI/CD pipeline to avoid that error on gitactions.

I see the empty `graphql_api/` folder is still there. Remember to try the above suggestion in next PR. Good to add an issue for tracking this too, if we cant get to it immediately

PR #607, opened 2023-06-30
Auth/endpoints

sijumoncy:
- #457 
- Changed DB schema for new table api_endpoints
- added POST, PUT, GET endpoints
- New table is to support api-permission-map

### DIFF
--- a/app/schema/schema_auth.py
+++ b/app/schema/schema_auth.py
@@ -440,3 +440,59 @@ class PermissionMapsResponse(BaseModel):
     """Response object of Permission map creation"""
     message:str = Field(..., example="Permission map created successfully")
     data: PermissionMapOut
+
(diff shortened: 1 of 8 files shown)


app/schema/schema_auth.py
```diff
@@ -439,4 +439,60 @@ class Config:
...
+            "example": {
+                "endpointId":100075,
+                "endpoint":'v2/access/permission-map',
+                "method": "GET",
+            }}
+
+class EndpointOut(BaseModel):
+    """auth endpoint output object"""
+    endpointId:int=None
+    endpoint:str
```
kavitharaju: Why do we need to keep the id field of out-object optional?

sijumoncy: fixed

kavitharaju (approved):

Issue #563, opened 2023-05-10
Last modified date showing incorrectly


AnnCherianA:
Issue description : The last modified date of a project is currently showing the date the user added a book to the project 
Expected result : The last modified date is supposed to show the date when the user last modified the translation in the project 

Link to the UI issue in gitlab : https://gitlab.bridgeconn.com/software/research/sign-language/sign-translation-assistant/-/issues/118

This issue will be picked up later for enhancement and is of low priority now, as discussed in the  ISL team meeting

AnnCherianA:
![image](https://github.com/Bridgeconn/vachan-api/assets/74647279/a1ffd72b-2e81-4e0f-9ae1-d5ba463f0aa4)

shimilgithub:
Shows `GMT` timezone instead of `IST`

PR #609, opened 2023-07-03
Updating timezone to Asia/Kolkata

shimilgithub:
- #563

### DIFF
--- a/app/db_models.py
+++ b/app/db_models.py
@@ -1,6 +1,7 @@
 ''' Defines SQL Alchemy models for each Database Table'''
 
 from enum import Enum
+from datetime import datetime
 from sqlalchemy import Column, Integer, String, JSON, ARRAY, Float, text
 from sqlalchemy import Boolean, ForeignKey, DateTime
 from sqlalchemy import UniqueConstraint, Index
@@ -10,9 +11,11 @@ from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.schema import Sequence
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.hybrid import hybrid_property
-
+from pytz import timezone
 from database import Base  # pylint: disable=import-error
 from custom_exceptions import GenericException # pylint: disable=import-error
+
+ist_time = datetime.now(timezone("Asia/Kolkata")).strftime('%Y-%m-%d %H:%M:%S')
 
 dynamicTables = {}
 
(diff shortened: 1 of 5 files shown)


app/db_models.py
```diff
@@ -10,10 +11,12 @@
 from sqlalchemy.schema import Sequence
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.hybrid import hybrid_property
-
+from pytz import timezone
 from database import Base  # pylint: disable=import-error
 from custom_exceptions import GenericException # pylint: disable=import-error
 
+ist_time = datetime.now(timezone("Asia/Kolkata")).strftime('%Y-%m-%d %H:%M:%S')
+
```
kavitharaju, line 19: This `ist_time` looks like a data variable to me rather than a function from this implementation. So what exactly would be the value that gets stored in this variable, the time app started or something like that or the time a certain API was called?

Adding a test to do the following could help figure it out.
1. In a testcase, create a project or a resource
2. In the same testcase, make a PUT call to update it. In the response, make sure there is at least a difference of fraction of seconds between the created and updated times.(Adding a wait of one or two second between these two calls will also help.

app/db_models.py
```diff
@@ -65,8 +69,9 @@ class License(Base): # pylint: disable=too-few-public-methods
     active = Column('active', Boolean)
     # metaData = Column('metadata', JSON)
     createdUser = Column('created_user', String)
+    createTime = Column('created_at',DateTime,default=ist_time)
     updatedUser = Column('last_updated_user', String)
-    updateTime = Column('last_updated_at', DateTime, onupdate=func.now())
+    updateTime = Column('last_updated_at', DateTime, onupdate=ist_time)
```
kavitharaju, line 75: I see you are setting the `onupdate` automatic update feature here. If, in the crud, we are passing a value for the "updateTime" field, there is chance that, that passed value will be taken and not this `ist_time` value. So, please make sure if the crud has any places where the time is passed from there, it is also in same timezone and format we are using here.

app/db_models.py
shimilgithub: `timezone` class in `pytz` package we use creates instances of different time zones. 
Also  `ist_time` will get updated upon API call.

Already there is a [test](https://github.com/Bridgeconn/vachan-api/blob/version-2/app/test/test_agmt_projects.py#L867C4-L867C4) on the mentioned scenario.

kavitharaju: Okay, if you have verified it, good

app/db_models.py
kavitharaju: I will wait for a response on this, before merging.

kavitharaju (approved):

Issue #623, opened 2023-07-07
Add parascripturals as a New Resource Type for Non-Bible Content


shimilgithub:
Add  a new resource type called `parascripturals` to support non-Bible content. By introducing this resource type, we can easily incorporate diverse non-Bible materials while removing the need for specific components related to infographics, Bible videos, and their associated endpoints and database tables.
[Discussions](https://github.com/Bridgeconn/vachan-api/discussions/575)

PR #632, opened 2023-07-12
new resource parascripturals

shimilgithub:
- #623 
- Has **database change**
- Included GET,POST,PUT,DELETE API calls of parascripturals
- Included testcases for parascripturals
- Removed  infographics and bible video endpoints, db tables and  related components in code


app/crud/contents_crud.py
```diff
@@ -328,25 +328,30 @@ def delete_dictionary(db_: Session, delitem : int,table_name = None,
...
-def get_infographics(db_:Session, source_name, book_code=None, title=None,**kwargs):
-    '''Fetches rows of infographics from the table specified by source_name'''
-    active = kwargs.get("active",True)
+def get_parascripturals(db_:Session, source_name, paratype=None, title=None,**kwargs):
+    '''Fetches rows of parascripturals from the table specified by source_name'''
+    search_word = kwargs.get("search_word",None)
     skip = kwargs.get("skip",0)
     limit = kwargs.get("limit",100)
-    infographic_id=kwargs.get("infographic_id",None)
+    parascript_id=kwargs.get("parascript_id",None)
```
kavitharaju: It would be required to have filtering options based on all available fields of parascripturals. 
- metadata
- reference: Not only for exact match, also for ranges. For example if table has a range, Gen 1:1 to Gen 1:20, then a query with Gen 1:5 should return the row
- content: Something like string contains. That is not exact match but sub string match or search option just on this field
- and any more fields we have included in parascripturals

app/crud/contents_crud.py
```diff
@@ -389,37 +389,31 @@ def upload_infographics(db_: Session, source_name, infographics, user_id=None):
...
                 f"title:{item.title}, "+\
                 f"not found for {source_name}")
-        if item.infographicLink:
-            row.infographicLink = item.infographicLink
-        if item.active is not None:
-            row.active = item.active
+        if item.description:
+            row.description = item.description
+        if item.content:
+            row.content = item.content
```
kavitharaju: We would need this unicode normalization for all text fields, where we may get values in other languages. Like title, description, content etc.

app/db_models.py
```diff
@@ -163,38 +162,21 @@ class Dictionary(): # pylint: disable=too-few-public-methods
...
+class Parascriptural():# pylint: disable=too-few-public-methods
+    '''Corresponds to the dynamically created parascripturals tables in vachan Db(postgres)'''
+    parascriptId = Column('parascript_id', Integer,
+        Sequence('parascriptural_id_seq', start=100001, increment=1), primary_key=True)
+    paratype = Column('paratype',String)
     title = Column('title', String)
-    infographicLink = Column('infographic_url', String)
-    active = Column('active', Boolean)
+    description = Column('description', String)
+    content = Column('content', String)
```
kavitharaju: Since we have active field in all other tables, good to keep it just for consistency

app/db_models.py
```diff
@@ -163,38 +162,21 @@ class Dictionary(): # pylint: disable=too-few-public-methods
...
-        return Column('book_id', Integer, ForeignKey('bible_books_look_up.book_id'))
-    @declared_attr
-    def book(cls): # pylint: disable=E0213
-        '''For modelling the book field in derived classes'''
-        return relationship(BibleBook)
+class Parascriptural():# pylint: disable=too-few-public-methods
+    '''Corresponds to the dynamically created parascripturals tables in vachan Db(postgres)'''
+    parascriptId = Column('parascript_id', Integer,
+        Sequence('parascriptural_id_seq', start=100001, increment=1), primary_key=True)
+    paratype = Column('paratype',String)
```
kavitharaju: This would need a better name. If `type` can't be used can we use either `content_type`(the older content type we had is going to be called resource_type, so that name is free I think) or `category` may be. This should be used alike in I/O objects(pydantic models, query params etc) , DB tables and preferably in code too.

app/db_models.py
```diff
@@ -163,38 +162,21 @@ class Dictionary(): # pylint: disable=too-few-public-methods
...
-        UniqueConstraint('book_id', 'title'),
+        UniqueConstraint('paratype', 'title'),
         {'extend_existing': True}
                      )
-
-class BibleVideo(): # pylint: disable=too-few-public-methods
-    '''Corresponds to the dynamically created bible videos tables in vachan Db(postgres)'''
-    bibleVideoId  = Column('biblevideo_id', Integer,
-        Sequence('biblevideo_id_seq', start=100001, increment=1), primary_key=True)
-    refIds  = Column('ref_ids', ARRAY(Integer))
```
kavitharaju: I wonder what would be the best way to store the references. 
1. In the deleted biblevideo code, i see we were listing all ref ids as an array. That is not very storage efficient if the ranges will be large.
2. In the current code I see you are using JSONB. That means queries with a verse in between can become super inefficient as we cant perform that query via SQL. but will have to fetch all data to memory and perform that filter some how in python. That is not gonna work out with large tables.
3. One design that comes to my mind is, in I/O objects keep it as a JSON with specified fields. In DB table keep two columns, ref_start and ref_end, which will have the integer presentation of the range boundaries. Now queries can be done using integer comparison of two fields uisng a between operation in SQL. 
Eg:
JSON in input may have fields representing `GEN 1:1` to `GEN 1:25`.
In DB table, `ref_start=1001001` and `ref_end=1001025`. (the bbcccvvv format, but integer values).
A query for verse GEN 1:7 could check if `ref_start <= 1001007 <= ref_end`

app/schema/schema_content.py
```diff
@@ -484,204 +484,84 @@ class DictionaryUpdateResponse(BaseModel):
...
-    bookCode : BookCodePattern
+class ParascriptEdit(BaseModel):
+    '''Input object of parascriptuals update'''
+    paratype: str
     title: str
-    infographicLink : AnyUrl = None
-    active: bool = None
+    description: str = None
+    content: str = None
+    link: AnyUrl = None
```
kavitharaju: Reference not included?

app/schema/schema_content.py
```diff
@@ -484,204 +484,84 @@ class DictionaryUpdateResponse(BaseModel):
...
+class ParascriptResponse(BaseModel):
+    '''Response object of parascripturals'''
+    parascriptId : int
+    paratype: str
     title: str
-    infographicLink : AnyUrl
-    active: bool
+    description: str = None
+    content: str = None
+    reference: dict = None
```
kavitharaju: reference's type should be a class defined separately specifying the inner fields

kavitharaju (changes requested):
Sharing my thoughts after a partial review. Tests and routers have not been reviewed.

app/db_models.py
shimilgithub: @kavitharaju  what would be the type of `reference` in I/O objects? Is that a separate `Reference class` with book,chapter,verse fields(as mentioned below)? or a json object having these fields(as mentioned above)

kavitharaju: In pydantic, define it as class, so that proper validation of all inner fields will be done. For the user, it would be a JSON with specific fields anyway, right?

app/crud/contents_crud.py
```diff
@@ -356,30 +409,51 @@ def get_infographics(db_:Session, resource_name, book_code=None, title=None,**kw
...
-def upload_infographics(db_: Session, resource_name, infographics, user_id=None):
-    '''Adds rows to the infographics table specified by resource_name'''
+def create_parascript_ref_id(db_:Session, book_code, chapter, verse):
+    '''Generate parascript ref start and end id'''
+    book_content = db_.query(db_models.BibleBook).filter(
+        db_models.BibleBook.bookCode == book_code.lower()).first()
+    book_id = book_content.bookId
+    if book_id is not None:
+        ref_id = (book_id * 100000) + (chapter * 1000) + verse
+    return ref_id
```
kavitharaju: I wonder if we already have this function somewhere. If not, a good place for this in the code base is in utils.py file so that other module can easily find it a use it.

app/crud/contents_crud.py
```diff
@@ -390,35 +464,45 @@ def upload_infographics(db_: Session, resource_name, infographics, user_id=None)
...
                 f"title:{item.title}, "+\
                 f"not found for {resource_name}")
-        if item.infographicLink:
-            row.infographicLink = item.infographicLink
+        if item.description:
+            row.description = item.description
+        if item.content:
+            row.content = item.content
+        if item.link:
+            row.link = item.link
```
kavitharaju: Need to make do the unicode normalization , stripping etc at PUT and POST functions as well. I see you do that in GET while filtering..

app/crud/contents_crud.py
```diff
@@ -390,35 +464,45 @@ def upload_infographics(db_: Session, resource_name, infographics, user_id=None)
...
+        if item.reference:
+            ref = jsonpickle.encode(item.reference)
+            ref = json.loads(ref)
+            ref = ref["py/state"]["__dict__"]
```
kavitharaju: What is happening here? if you are trying to convert a python dictionary into Pydantic or SQLAlchemy or other kinds of classes, there would be utils function that does that. See if you can either reuse it or re-plicate the logic.

kavitharaju: If its otherway round, most likely these classes will have `.__dict__()` fucntion to convert them to dictionary.

app/db_models.py
```diff
@@ -163,38 +162,24 @@ class Dictionary(): # pylint: disable=too-few-public-methods
...
+    parascriptId = Column('parascript_id', Integer,
+        Sequence('parascriptural_id_seq', start=100001, increment=1), primary_key=True)
+    category = Column('category',String)
     title = Column('title', String)
-    infographicLink = Column('infographic_url', String)
+    description = Column('description', String)
+    content = Column('content', String)
+    reference = Column('reference', JSONB)
+    refStart = Column('ref_start', Integer)
+    refEnd = Column('ref_end', Integer)
```
kavitharaju, line 175: I see you are replicating the same data in JSON and integer format... Good :+1:

app/schema/schema_content.py
```diff
@@ -115,6 +115,49 @@ class Config:
...
+        orm_mode = True
+        '''display example value in API documentation'''
+        schema_extra = {
+            "example": {
+                # "bible": "hi_IRV_5_bible",
+                "bookStart": "mat",
+                "chapterStart": 1,
+                "verseStart": 12,
+                "bookEnd": "luk",
+                "chapterEnd": 10,
```
kavitharaju: In an example in router module, I saw you were showing interger value expected in `bookStart`. I hope we are expecting it to be book code in all IO consistantly..

app/schema/schema_content.py
```diff
@@ -484,200 +527,92 @@ class DictionaryUpdateResponse(BaseModel):
...
+    category: str
     title: str
-    references: List[Reference] = []
-    videoLink: AnyUrl
-    description: str
-    series: str
-    active: bool
+    description: str = None
+    content: str = None
+    reference: ParascriptReference = None
```
kavitharaju: This `ParascriptReference` could be given a more generic name like `Reference` and used in other palces too... I can totally see us using the same approach in commentaries soon(ie, if others also like this)

shimilgithub: We already have schema named  [`Reference`](https://github.com/Bridgeconn/vachan-api/blob/API-refactor/app/schema/schema_content.py#L97) . What about `Scripture` or `Annotation`?

kavitharaju: Oh.. what we want is changing that existing Reference into the new format, with across chapter and across book scope. We only need one, that is to be used consistently in all places. 
can you try to do that without hurting things too much? Keep the newly added fields as optional and see if we can rename the existing fields as needed and make subsequent code changes without too much wreckage

app/crud/contents_crud.py
shimilgithub: does  unicode  normalization is enough for  `title` and `content` fields only?

kavitharaju (approved):

Issue #608, opened 2023-07-03
Update Terminology: Change "source" and "source_name" to "resource" and "resource_name"


shimilgithub:
The terms `source` and `source_name` needs to be modified as  `resource` and `resource_name` respectively in the codebase.

PR #634, opened 2023-07-14
change terminology

Joel-Joseph-George:
#608


app/crud/nlp_crud.py
```diff
@@ -126,19 +126,19 @@ def replace_bulk_tokens(db_, sentence_list, token_translations, src_code, trg_co
     '''Substitute tokens with provided trabslations and get updated drafts, draftMetas
     and add knowledge to translation memory'''
     use_data = kwargs.get("use_data_for_learning",True)
-    source = db_.query(db_models.Language).filter(
+    resource = db_.query(db_models.Language).filter(
```
shimilgithub: I think in `translation` module, terminology change is not required unless we are referring the `resources` table. @kavitharaju  can you please confirm this?

Joel-Joseph-George: ohh. Just confirm it and will make the changes as per need

kavitharaju: Yes, normally `source` in translation module refers to the source of translation as in source ad target. We should not change that to resource

app/test/test_agmt_translation2.py
shimilgithub, line 1: Still you  `resource` and related terms  is used in translation test. it has to be `source` . eg:source_sentences

app/test/test_smast_translation2.py
```diff
@@ -20,7 +20,7 @@
     "targetLanguageCode": "ml"
 }
 
-source_sentences = [
+resource_sentences = [
```
shimilgithub: should be `source_sentences`

app/test/test_smast_translation2.py
```diff
@@ -152,13 +152,13 @@ def test_draft_update_negative():
         headers=headers_auth, json=put_data)
     assert resp.json()['error'] == "Requested Content Not Available"
 
-    # upload source and repeat last action
-    put_data_source = {
+    # upload resource and repeat last action
+    put_data_resource = {
```
shimilgithub: should be `put_data_source`

shimilgithub (changes requested):
few changes required in ` test_agmt_translation2` and `test_smast_translation2` files as mentioned. Also remove the print statements you have kept in between.

shimilgithub (approved):

PR #639, opened 2023-07-21
Auth/permission map

sijumoncy:
- #457 
- Added Get endpoint 
- modified POST and PUT endpoint based on new table for endpoints 
- All changes of PR 
- All changes in #584 added in this PR


kavitharaju (approved):

kavitharaju:
Merging this with the error in initial setup.
As next pan is to move this module to a new repo, there we may not have conflicts with package dependencies as only those packages need by auth module needs to be maintained.

Issue #620, opened 2023-07-07
Update Terminology: Change dictionaries -> vocabularies


shimilgithub:
The term `dictionaries` needs to be modified as `vocabularies` in the codebase.
[Discussions](https://github.com/Bridgeconn/vachan-api/discussions/575)

PR #640, opened 2023-07-24
changes in terminology,dictionary to vocabulary

Joel-Joseph-George:
#620


app/test/test_dictionaries.py
```diff
@@ -1,4 +1,4 @@
-'''Test cases for dictionary related APIs'''
+'''Test cases for vocabulary related APIs'''
```
kavitharaju, line 1: Are we not changing the file name too?

kavitharaju (commented):
Looks good to me, other than the one point mentioned. Consider I approved. Will let @shimilgithub do the final review after the change

app/test/test_dictionaries.py
Joel-Joseph-George: ohh i think i forgot about the file name.. i can change it right now

shimilgithub (approved):

Issue #438, opened 2022-10-14
Need regular users for Vachan Admin App


kavitharaju:
For the Vachan-admin app that is going to serve as both the content managment tool and content dashboard for the VE contents, we need to have capability to register user.

As per the current design, we expect only the user "vachan-admin" to access APIs via the tool. There should be regualr users also who only have read access on the content but be able to view all things.

PR #643, opened 2023-07-26
New App:VachanContentDashboard with admin and user

shimilgithub:
- #438 
- Introduced new app `VachanContentDahboard` with two users `VachanContentAdmin` and `VachanContentViewer`
- `VachanContentAdmin` has same previlleges of `VachanAdmin`
- `VachanContentViewer` only has `read access` to content and metacontent.

### DIFF
--- a/app/auth/api-permissions.csv
+++ b/app/auth/api-permissions.csv
@@ -46,7 +46,7 @@ Endpoints,Method,RequestApp,FilterResults,ResourceType,Permission
 /v2/sources,GET,SanketMAST,True,None,refer-for-translation
 /v2/sources,GET,Vachan-online or vachan-app,True,None,view-on-web
 /v2/sources,GET,VachanAdmin,True,None,read-via-vachanadmin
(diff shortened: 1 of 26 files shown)


kavitharaju (dismissed):

app/auth/api-permissions.csv
```diff
@@ -42,6 +46,7 @@ Endpoints,Method,RequestApp,FilterResults,ResourceType,Permission
 /v2/sources,GET,SanketMAST,True,None,refer-for-translation
 /v2/sources,GET,Vachan-online or vachan-app,True,None,view-on-web
 /v2/sources,GET,VachanAdmin,True,None,read-via-vachanadmin
+/v2/sources,GET,VachanContentDashboard,False,None,read-via-vachancontentdashboard
```
sijumoncy: @shimilgithub ,
Do we need this filter to be TRUE , above VA with same permission-map have TRUE value ?

shimilgithub: Yes, It has to be `True`

sijumoncy (approved):

Issue #625, opened 2023-07-07
Add a new resource type for Sign Bible Videos


shimilgithub:
A new resource type, `Sign Bible Videos` has to be added to cater specifically to sign language interpretations of the Bible.
[Discussions](https://github.com/Bridgeconn/vachan-api/discussions/575)

PR #652, opened 2023-07-28
New resource type sign bible videos

shimilgithub:
- #625 
- **includes database change**
- Included GET,POST,PUT,DELETE API calls of sign bible video
- Included testcases for sign bible video
- handled cases where `verse start number` and `verse end number` are absent
- handles cases where  `chapter number` is absent or invalid
-  Improved way of handling reference in #623


app/crud/contents_crud.py
```diff
@@ -481,10 +495,16 @@ def update_parascripturals(db_: Session, resource_name, parascripturals, user_id
...
             ref = item.reference.__dict__
-            ref_start = \
-                utils.create_parascript_ref_id(db_,ref['book'],ref['chapter'],ref['verseNumber'])
-            ref_end  = \
-                utils.create_parascript_ref_id(db_,ref['bookEnd'],ref['chapterEnd'],ref['verseEnd'])
+            if ref['verseNumber'] is not None:
+                ref_start = utils.create_decimal_ref_id(
+                    db_,ref['book'],ref['chapter'],ref['verseNumber'])
+            else:
+                ref_start = utils.create_decimal_ref_id(db_,ref['book'],ref['chapter'],0)
```
kavitharaju, line 505: I think the `create_decimal_ref()` function itself would be the right place to handle the `None` to `000` converstion. We may need to repeat this from different calling functions otherwise.

Also good to handle the following cases too
1. chapter is None and verse is None => chapter=0 and verse=0
2. chapter is None and verse in not None => raise 422 error

kavitharaju: The the ref_end handling could also be done centrally in the utils function `create_decimal_ref_id()`, it seems

app/crud/contents_crud.py
```diff
@@ -519,6 +540,184 @@ def delete_parascriptural(db_: Session, delitem: int,table_name = None,\
...
+            query = query.filter(key_match | value_match)
+    if search_word:
+        search_pattern = " & ".join(re.findall(r'\w+', search_word))
+        search_pattern += ":*"
+        query = query.filter(text("to_tsvector('simple', title || ' ' ||"+\
+            " signvideo_id || ' ' || "+\
+            " description || ' ' || "+\
+            " link || ' ' || "+\
+            " reference || ' ' || "+\
+            "jsonb_to_tsvector('simple', metadata, '[\"string\", \"numeric\"]') || ' ')"+\
```
kavitharaju: At the db level `reference` is also a jsonb field like metadata right? Wont it need some kind of handling?

app/crud/contents_crud.py
```diff
@@ -519,6 +540,184 @@ def delete_parascriptural(db_: Session, delitem: int,table_name = None,\
...
+                #setting verseEnd to 999 if its not present
+                ref_end   = utils.create_decimal_ref_id(db_,ref['bookEnd'],ref['chapterEnd'],999)
+                ref['verseEnd'] = 999
+        else:
+            ref = None
+            ref_end = None
+            ref_start = None
+        row = model_cls(
+            title = utils.normalize_unicode(item.title.strip()),
+            description = item.description,
```
kavitharaju: description could also be unicode normalized

app/crud/contents_crud.py
```diff
@@ -519,6 +540,184 @@ def delete_parascriptural(db_: Session, delitem: int,table_name = None,\
...
+    for item in signvideos:
+        row = db_.query(model_cls).filter(
+            model_cls.signVideoId == item.signVideoId).first()
+        if not row:
+            raise NotAvailableException(f"Sign Bible Video row with id:{item.signVideoId}, "+\
+                f"not found for {resource_name}")
+        if item.title:
+            row.title = item.title
+        if item.description:
+            row.description = item.description
```
kavitharaju: both these could be unicode normalized

app/crud/contents_crud.py
```diff
@@ -519,6 +540,184 @@ def delete_parascriptural(db_: Session, delitem: int,table_name = None,\
...
+            row.refStart=ref_start
+            row.refEnd=ref_end
+        if item.metaData:
+            row.metaData = item.metaData
+        if item.active is not None:
+            row.active = item.active
+        db_.flush()
+        db_content.append(row)
+    resource_db_content.updatedUser = user_id
+    resource_db_content.updateTime = datetime.now(ist_timezone).strftime('%Y-%m-%d %H:%M:%S')
```
kavitharaju, line 704: I see we are not passing the created time in POST api. If it is being filled from DB, is it going to be the same format as this?

app/crud/contents_crud.py
```diff
@@ -519,6 +540,184 @@ def delete_parascriptural(db_: Session, delitem: int,table_name = None,\
...
+    response = {
+        'db_content':db_content,
+        'resource_content':resource_db_content
+        }
+    return response
+
+def delete_sign_bible_videos(db_: Session, delitem: int,table_name = None,\
+    resource_name=None,user_id=None):
+    '''delete particular item from sign bible video, selected via resourceName and signvideo id'''
```
kavitharaju, line 713: Are we not adding the deleted content to deleted_items table?

kavitharaju (changes requested):

app/crud/contents_crud.py
shimilgithub: But we are passing only `verse` in utils function. It can be `verseNumber`(000) or `verseEnd`(999).   Value needs to be passed according to the context(if verseNumber is absent `verse=000`,if verseEnd is absent `verse=999`. 

Else we may need to:
1.pass both `verseNumber` and `verseEnd` in utils function. OR
2. write two different functions for `verseNumber` and `verseEnd`

app/crud/contents_crud.py
shimilgithub: `createTime` alos set in the same [timezone](https://github.com/Bridgeconn/vachan-api/pull/652/files#diff-e46e66fe2bf95836154ce382dabbdefbe2f05b038396b07de81f9d3daf48d133R202) in pydantic model

app/crud/contents_crud.py
shimilgithub: Yes. we are adding it [here](https://github.com/Bridgeconn/vachan-api/pull/652/files#diff-71693af643ae545f712c729ad117046a7a8b43f319cd78a43cb30552e867cab8R1220)

app/crud/contents_crud.py
kavitharaju: Oh I see... Then I wont press for this change.

But as I already see similar kind of code being repeated in this PR itself and I expect its gonna be needed several more times, if you think the change is good, then go for it. In that case I like option 1 you said, keeping end params optional.

app/crud/contents_crud.py
kavitharaju: Oh yes, In the router function. Good. I missed it.

app/crud/contents_crud.py
shimilgithub: keeping it as same function since it disturbs `filter by single verse`

app/test/test_sign_bible_videos.py
```diff
@@ -300,7 +299,7 @@ def test_searching():
     assert found
 
     # searching with reference:partial match - positive test
-    response = client.get(UNIT_URL+resource_name+"?search_word=chapter:2",headers = headers_auth)
+    response = client.get(UNIT_URL+resource_name+"?search_word=MA",headers = headers_auth)
```
kavitharaju: I wonder if searching for MAT 2 would find us what we need here. I hope it would

kavitharaju (approved):

Issue #442, opened 2022-10-21
Need to distinguish translation projects based on Apps


kavitharaju:
Now there is only one translation project working using our translation APIs. That is the sign project. But later if new apps come then we have a need to distinguish which project belongs to which app, also specify if there are projects that are sharable across apps.

For instance the sign app works with its own specific logic that each target word(token) of the translatio would a valid sign token with its dictionary entry. If this app shows a text project in its list of projects and tries to display its contents via the sign UI, it will never work. So there should a way to say only sign project are to loaded on sign app. For that we need to keep the record of which app created which project, even if the rest of translation logic are shareable.

I suggest either a new column in the translation projects table or if we don't want to change APIs and DB schema, a field in project metadata.

joelthe1:
I think having a field, say `compatible_with`, that is a list of strings that name the apps (based on some standard app names in the DB) might be enough.

PR #665, opened 2023-08-09
Distinguish translation projects based on Apps

shimilgithub:
- #442 
- **has database change**
- **Moved `project_id` from body param to query param in PUT projects and PUT user APIs**
- App compatibility check added in following APIs
   - PUT ,POST,DELETE of `/v2/text/translate/token-based/projects`
   - PUT,POST,DELETE of `/v2/text/translate/token-based/project/user`
   - GET,PUT of `/v2/text/translate/token-based/project/tokens`
   - GET of `/v2/text/translate/token-based/project/token-translations`
   - GET of `/v2/text/translate/token-based/project/token-sentences`
   - GET,PUT of `/v2/text/translate/token-based/project/draft`
   - GET,DELETE of `/v2/text/translate/token-based/project/sentences`
   - GET `/v2/text/translate/token-based/project/progress`
   - GET `/v2/text/translate/token-based/project/versification`
   - PUT `/v2/text/translate/token-based/suggestions`

### DIFF
--- a/app/crud/projects_crud.py
+++ b/app/crud/projects_crud.py
@@ -9,6 +9,7 @@ from functools import wraps
 from pytz import timezone
 from sqlalchemy import or_
 from sqlalchemy.orm import Session
+from sqlalchemy import text
 from sqlalchemy.orm.attributes import flag_modified
 from fastapi import HTTPException
 import db_models
@@ -37,7 +38,7 @@ def create_translation_project(db_:Session, project, user_id=None,app=None):
     if project.punctuations:
         meta['punctuations'] = project.punctuations
     if project.compatibleWith is None:
-        project.compatibleWith = app
+        project.compatibleWith = [app]
     db_content = db_models.TranslationProject(
         projectName=utils.normalize_unicode(project.projectName),
         source_lang_id=source.languageId,
@@ -196,20 +197,67 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
     return project_row
 
 def check_app_compatibility_decorator(func):#pylint:disable=too-many-statements
-    """Decorator function for ato check app compatibility"""
+    """Decorator function for to check app compatibility"""
     @wraps(func)
-    async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-statements, too-many-locals
+    async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-statements
         request = kwargs.get('request')
-        body = await request.body()
-        json_string = body.decode()
-        parsed_data = json.loads(json_string)
-        if 'compatibleWith' in parsed_data and parsed_data['compatibleWith'] is not None:
-            compatible_with = parsed_data['compatibleWith']
+        db_ = kwargs.get("db_")
+        endpoint = request.url.path
+        method = request.method
+        query_params = request.query_params
+        compatible_with = []
+
+        if method == 'POST' and endpoint == '/v2/text/translate/token-based/projects':
+            #POST method to create project
+            #assign compatibility_with to userinput(if present),else assign to header app
+            body = await request.body()
+            json_string = body.decode()
+            parsed_data = json.loads(json_string)
+            if 'compatibleWith' in parsed_data and parsed_data['compatibleWith'] is not None:
+                compatible_with = parsed_data['compatibleWith']
+            else:
+                compatible_with = request.headers['app']
+        elif method == 'POST':
+            # POST method for other translation endpoints
+            # assign compatibility_with to respective db field value
+            project_obj = db_.query(db_models.TranslationProject).first()
+            project_row = project_obj.compatibleWith
+            if project_row is not None:
+                compatible_with = project_row
+        elif method == 'PUT':
+            #For PUT method: assign compatibility_with to respective db field value
+            body = await request.body()
+            if len(body) != 0:
+                json_string = body.decode()
+                parsed_data = json.loads(json_string)
+                if 'projectId' in parsed_data:
+                    project_id = parsed_data['projectId']
+                elif 'project_id' in parsed_data:
+                    project_id = parsed_data['project_id']
+                else:
+                    project_id = query_params['project_id']
+            else:
+                project_id = query_params['project_id']
+            project_obj = db_.query(db_models.TranslationProject).get(project_id)
+            if project_obj is not None:
+                compatible_with = project_obj.compatibleWith
+        elif method == 'DELETE':
+            # assign compatibility_with to respective db field value
+            project_id = query_params['project_id']
+            project_obj = db_.query(db_models.TranslationProject).get(project_id )
+            if project_obj is not None:
+                compatible_with = project_obj.compatibleWith
         else:
-            compatible_with = request.headers['app']
+            project_obj = db_.query(db_models.TranslationProject).first()
+            project_row = project_obj.compatibleWith
+            if project_row is not None:
+                compatible_with = project_row
         if 'app' in request.headers:
             client_app = request.headers['app']
-            if client_app != compatible_with:
+            if len(compatible_with) ==0:
+                raise HTTPException(status_code=404,detail = \
+                    f"Project with id, {project_id}, not present")
+            if client_app not in compatible_with:
                 raise HTTPException(status_code=403, detail="Incompatible app")
         else:
             raise HTTPException(status_code=403, detail="Incompatible app")
@@ -240,8 +288,10 @@ def get_translation_projects(db_:Session, project_name=None, source_language=Non
         if not target:
             raise NotAvailableException(f"Language, {target_language}, not found")
         query = query.filter(db_models.TranslationProject.target_lang_id == target.languageId)
-    if compatible_with:
-        query = query.filter(db_models.TranslationProject.compatibleWith == compatible_with)
+    if compatible_with is not None:
+        query = query.filter(text(
+            "ARRAY[:compatible_with]::text[] <@ translation_projects.compatible_with").bindparams(
+            compatible_with=compatible_with))
     if user_id:
         query = query.filter(db_models.TranslationProject.users.any(userId=user_id))
     query = query.filter(db_models.TranslationProject.active == active)
--- a/app/crud/projects_crud.py
+++ b/app/crud/projects_crud.py
@@ -195,8 +195,8 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
     # db_.commit()
     # db_.refresh(project_row)
     return project_row
-
-def check_app_compatibility_decorator(func):#pylint:disable=too-many-statements,duplicate-code
+# pylint: disable=duplicate-code
+def check_app_compatibility_decorator(func):#pylint:disable=too-many-statements
     """Decorator function for to check app compatibility"""
     @wraps(func)
     async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-statements
@@ -263,6 +263,7 @@ def check_app_compatibility_decorator(func):#pylint:disable=too-many-statements,
             raise HTTPException(status_code=403, detail="Incompatible app")
         return await func(*args, **kwargs)
     return wrapper
+# pylint: enable=duplicate-code
 
 def get_translation_projects(db_:Session, project_name=None, source_language=None,
     target_language=None, **kwargs):
--- a/app/crud/projects_crud.py
+++ b/app/crud/projects_crud.py
@@ -215,7 +215,7 @@ def check_app_compatibility_decorator(func):#pylint:disable=too-many-statements
                 project_id = parsed_data['projectId']
             elif 'project_id' in parsed_data:
                 project_id = parsed_data['project_id']
-            elif 'project_id' in query_params:
+            if 'project_id' in query_params:
                 project_id = query_params['project_id']
         else:
             project_id = query_params['project_id']
@@ -226,8 +226,6 @@ def check_app_compatibility_decorator(func):#pylint:disable=too-many-statements
         else:
             if 'compatibleWith' in parsed_data and parsed_data['compatibleWith'] is not None:
                 compatible_with = parsed_data['compatibleWith']
-            else:
-                compatible_with = request.headers['app']
         if 'app' in request.headers:
             client_app = request.headers['app']
             if len(compatible_with) ==0:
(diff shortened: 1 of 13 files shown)


app/crud/projects_crud.py
```diff
@@ -187,10 +195,32 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
     # db_.refresh(project_row)
     return project_row
 
+def check_app_compatibility_decorator(func):#pylint:disable=too-many-statements
+    """Decorator function for ato check app compatibility"""
```
kavitharaju: Does this function have too many statements? I think this comment could be safely removed

app/crud/projects_crud.py
```diff
@@ -187,10 +195,32 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
+    @wraps(func)
+    async def wrapper(*args, **kwargs):#pylint: disable=too-many-branches,too-many-statements, too-many-locals
+        request = kwargs.get('request')
+        body = await request.body()
+        json_string = body.decode()
+        parsed_data = json.loads(json_string)
```
kavitharaju: I dont know if we need to do all this to get the arguments. If its body parameter, it is a kwarg to the function right? It might be accessible from the kwargs.get(<body param name>)

app/crud/projects_crud.py
```diff
@@ -187,10 +195,32 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
+        if 'compatibleWith' in parsed_data and parsed_data['compatibleWith'] is not None:
+            compatible_with = parsed_data['compatibleWith']
+        else:
+            compatible_with = request.headers['app']
+        if 'app' in request.headers:
+            client_app = request.headers['app']
+            if client_app != compatible_with:
+                raise HTTPException(status_code=403, detail="Incompatible app")
+        else:
+            raise HTTPException(status_code=403, detail="Incompatible app")
```
kavitharaju: 1. Taking `app` value from header and then checking against it is pointless.
2. In 207, I assume we expect the `compatible_with` data to come as input of the function. I hope that input is not coming from user and it is coming from DB. If it is coming from user, it should be treated here as `client_app` and values for `compatible_app` should come from DB.
3. `client_app != compatible_with` should actually be `client_app in compatible_with`. Meaning we need to have a list of apps in `compatible_with`

app/routers/translation_apis.py
```diff
@@ -27,6 +26,7 @@ async def get_projects(request: Request,
     source_language:schemas.LangCodePattern=Query(None,example='en'),
     target_language:schemas.LangCodePattern=Query(None,example='ml'),
     active:bool=True, user_id:str=Query(None),
+    compatible_with: schema_auth.App = Query(None,example="SanketMAST"),
```
kavitharaju: Doing this in `get_project()` looks good. One suggestion is , to set it to requesting app, from header if None. Also to make it list.

kavitharaju (changes requested):
Converting the field from str to list is going to have impact through out the code, like DB, pydantic schema etc. But mentioned it only in one place

app/routers/translation_apis.py
```diff
@@ -278,6 +284,7 @@ async def get_draft(request: Request,project_id:int=Query(...,example="1022004")
     415: {"model": schemas.ErrorResponse},404: {"model": schemas.ErrorResponse}},
     tags=['Project-Based-Translation'])
 @get_auth_access_check_decorator
+@check_app_compatibility_decorator
 async def update_draft(request: Request,project_id:int=Query(...,example="1022004"),
```
kavitharaju, line 299: Are we using this check for all APIs in translation projects or just a few?

app/crud/projects_crud.py
shimilgithub: > 1. Taking `app` value from header and then checking against it is pointless.
> 2. In 207, I assume we expect the `compatible_with` data to come as input of the function. I hope that input is not coming from user and it is coming from DB. If it is coming from user, it should be treated here as `client_app` and values for `compatible_app` should come from DB.
> 3. `client_app != compatible_with` should actually be `client_app in compatible_with`. Meaning we need to have a list of apps in `compatible_with`

Point 2 is not clear for me.

`compatible_with` in line 207 comes from user input.

Does taking "compatible_with" value from db will be same as that of the app value in request?Because we are storing in the db only if "compatible_with" value from user input matches "app" value in request.

This is how I implemented.
Initially in POST:
1. we are expecting compatible_with value from user input,if not there, it is set as app in the header
2. If compatible_with from user input = app in the request -> test pass, else fail

app/routers/translation_apis.py
shimilgithub: check is given for follow

> Are we using this check for all APIs in translation projects or just a few?

check is done for following APIs
PUT and POST /v2/text/translate/token-based/projects and filtering in GET
PUT /v2/text/translate/token-based/project/tokens
PUT /v2/text/translate/token-based/project/token-sentences
PUT /v2/text/translate/token-based/project/draft

kavitharaju: I can see total 16 APIs there, without counting GET projects and POST project. All the rest can have this check, right?

app/crud/projects_crud.py
shimilgithub: can i get confirmation on this?

kavitharaju: How you are setting it in POST projects is correct. That is, expect user to provide the value to you, if not provided take from header and set that value inn DB.

Now what are we doing this for? For making sure projects are served to only app's that they are compatible with , right? So the details we need for them are two
1. Nature of the project data. Which we should be getting from saved value in DB. That is what i would assign for the "compatible_with" variable in your decorator. So this should not come from user, but from DB, for all apis calls except the one to create it for the first time.
2. The nature of the app making the API call. Which should be obtained from the API calls header. This is what should be called as `client_app` in your code.

More over the `compatible_with` in DB and also here should be a list, not just single string, to accommodate such needs in future.

shimilgithub: @kavitharaju I am stucked in below scenario.
```
client_app(app in header) =  SanketMAST
1. POST -> compatiblileWith(user input) = ['SanketMAST','Autographa']
   project created
   db_app(app value stored in db) = ['SanketMAST','Autographa']
2. POST  -> compatible_with = ["VachanAdmin']
   raises error "incomptible app"
3. PUT  -> compatibleWith(user input) = ['Vachanadmin']
   db app = ['SanketMAST','Autographa']
   client_app = ['SanketMAST']
   PROJECT UPDATED since db_app = client app
   Now db_app - ['VachanAdmin']
```
After step 3, I am unable to perform any  other operations(PUT/DELETE) since client_app and db_app are different.

kavitharaju: In my mind I imagine the PUT in step 3 to be used like this
1. Is we want to add one more app to the existing list, we give the full list ['SanketMAST','Autographa', 'Vachanadmin']
2. If we want to remove one app from the existing list we provide a list without it, but including the rest of the apps already present in the list ['SanketMAST']

app/crud/projects_crud.py
```diff
@@ -194,27 +195,75 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
+        if method == 'POST' and endpoint == '/v2/text/translate/token-based/projects':
+            #POST method to create project
+            #assign compatibility_with to userinput(if present),else assign to header app
+            body = await request.body()
+            json_string = body.decode()
+            parsed_data = json.loads(json_string)
+            if 'compatibleWith' in parsed_data and parsed_data['compatibleWith'] is not None:
+                compatible_with = parsed_data['compatibleWith']
+            else:
+                compatible_with = request.headers['app']
```
kavitharaju: Do we need this `if` block at all? Wouldn't it work if we just not call the decorator from this API route(`POST /v2/text/translate/token-based/projects`)?

app/crud/projects_crud.py
```diff
@@ -194,27 +195,75 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
+        elif method == 'POST':
+            # POST method for other translation endpoints
+            # assign compatibility_with to respective db field value
+            project_obj = db_.query(db_models.TranslationProject).first()
+            project_row = project_obj.compatibleWith
```
kavitharaju: I see you are making a query with not filter(WHERE clause), which means you are always going to get the first project in the table? is that how you intended this to work? Wont we have project_id specified in POST apis?

app/crud/projects_crud.py
```diff
@@ -194,27 +195,75 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
+            if project_row is not None:
+                compatible_with = project_row
```
kavitharaju: In other `elif` blocks I see you are setting the value like this
```
 compatible_with = project_obj.compatibleWith
```
Is using the project row here working properly where as in other places you have the specific field?

app/crud/projects_crud.py
```diff
@@ -194,27 +195,75 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
+            project_id = query_params['project_id']
+            project_obj = db_.query(db_models.TranslationProject).get(project_id )
+            if project_obj is not None:
+                compatible_with = project_obj.compatibleWith
         else:
-            compatible_with = request.headers['app']
+            project_obj = db_.query(db_models.TranslationProject).first()
+            project_row = project_obj.compatibleWith
+            if project_row is not None:
+                compatible_with = project_row
```
kavitharaju: Same 2 questions asked above
1. Why the first project from table?
2. Why ` compatible_with = project_row` and not ` compatible_with = project_row.compatibleWith`?

kavitharaju: Also I wonder if we really need these different if-elif ladder, based on METHODs, wouldn't the code within `PUT` support call cases where project_id is used?

app/crud/projects_crud.py
```diff
@@ -194,27 +195,75 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
         if 'app' in request.headers:
             client_app = request.headers['app']
-            if client_app != compatible_with:
+            if len(compatible_with) ==0:
+                raise HTTPException(status_code=404,detail = \
+                    f"Project with id, {project_id}, not present")
+            if client_app not in compatible_with:
                 raise HTTPException(status_code=403, detail="Incompatible app")
         else:
             raise HTTPException(status_code=403, detail="Incompatible app")
```
kavitharaju: Good to define a custom_exception for this or use one of existing like `PermissionException` instead of the HTTPException.

app/crud/projects_crud.py
```diff
@@ -194,27 +195,75 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
+                json_string = body.decode()
+                parsed_data = json.loads(json_string)
+                if 'projectId' in parsed_data:
+                    project_id = parsed_data['projectId']
+                elif 'project_id' in parsed_data:
+                    project_id = parsed_data['project_id']
+                else:
+                    project_id = query_params['project_id']
+            else:
+                project_id = query_params['project_id']
```
kavitharaju: I see that all APIs, other than `PUT projects` and `PUT project/user` uses `project_id` as query parameter consistantly. Would it be good to change these two endpoints also to follow that pattern through out. Then our APIs, aswell as code snippets like this will be much more "neat". What do you think?

app/routers/translation_apis.py
```diff
@@ -26,14 +26,17 @@ async def get_projects(request: Request,
...
     skip: int=Query(0, ge=0), limit: int=Query(100, ge=0),
     user_details =Depends(get_user_or_none), db_:Session=Depends(get_db),
     filtering_required=Depends(AddHiddenInput(value=True))):
     '''Fetches the list of proejct and their details'''
     log.info('In get_projects')
     log.debug('project_name: %s, source_language:%s, target_language:%s,\
         active:%s, user_id:%s',project_name, source_language, target_language, active, user_id)
+    if compatible_with is None:
+        app = request.headers['app']
+        compatible_with = app
```
kavitharaju: Is this working properly? Dont we need `compatible_with = [app]`, as it is list?

kavitharaju (changes requested):

app/crud/projects_crud.py
shimilgithub: before line 225,`project_row` is assigned to  `project_obj.compatibleWith`

app/crud/projects_crud.py
```diff
@@ -195,72 +195,48 @@ def update_translation_project(db_:Session, project_obj, user_id=None):
...
                 compatible_with = project_obj.compatibleWith
         else:
-            project_obj = db_.query(db_models.TranslationProject).first()
-            project_row = project_obj.compatibleWith
-            if project_row is not None:
-                compatible_with = project_row
+            if 'compatibleWith' in parsed_data and parsed_data['compatibleWith'] is not None:
+                compatible_with = parsed_data['compatibleWith']
+            else:
+                compatible_with = request.headers['app']
```
kavitharaju: Here again in the else part I fear you are taking the value of `compatible_with` from user(either header or args) and in next step comparing it against itself.
Why do we need a an else for project_id is not None. Which senario will reslut in that need where we dont have project information to check against?

shimilgithub: @kavitharaju  `else`  part will execute only if `project` id is `None` and there is no `compatibleWith`(or compatibleWith is empty) field in the  `body param` of request.

This will be happening only for POST project APIs without compatibleWith field in the request body.

- PUT and DELETE  and POST user will pass through the condition at line 222 since `project_id` is mandatory there.
-  In POST, project
    -  if we are mentioning `compatibleWith` field in the request body,it will go to condition at line 227
    - if `compatibleWith` field not mentions, whatever in the app header will be assigned to it.(which is the else part at line 229)

kavitharaju: I thought we are not calling this check from POST project API. I dont see why that would make sense, to do this check when a new project is created fresh.

shimilgithub: Won't it be good to have the header app in `compatibleWith` list initially?

kavitharaju: Alright, I leave it to you, even though I dont quite see the point in doing a check like this

- Someone makes an API request from app XXX, to create a project
- We take that value, XXX,  from header or input  at line 228 or 230 and checks it with the same value we take from header or input at line 236.

I would say not include this decorator for that POST project API router function. But If you want to use it and follow consistency, it good. You can do it if you see value in it

shimilgithub: ok. will make changes

shimilgithub: ok. will make changes

kavitharaju (approved):

Issue #631, opened 2023-07-11
FastAPI Upgrade


kavitharaju:
- Fast API newer version available https://github.com/tiangolo/fastapi/releases/tag/0.100.0
- Other packages also have newer versions available: starllete, pydantic, pytest, SQLAlchemy, psycopg2
- Need to check which all versions are compatible with each other and bump to the latest possible versions

shimilgithub:
https://github.com/tiangolo/fastapi/releases

PR #671, opened 2023-08-21
UPGRADATION: FAST API - 0.1.10.1 __No changes in pydantic

Joel-Joseph-George:
#631

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 SQLAlchemy==1.4.41
 python-gitlab==3.9.0
-fastapi[all]==0.95.0
+fastapi==0.101.1
+pydantic==1.10.11
 pygtrie==2.5.0
 nltk==3.7
 numpy==1.23.3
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1,6 +1,10 @@
 SQLAlchemy==1.4.41
 python-gitlab==3.9.0
-fastapi[all]==0.95.0
+fastapi==0.101.1
+pydantic==1.10.11
+httpx==0.24.1
+python-multipart==0.0.6
+Jinja2==3.1.2
```
shimilgithub, line 7: @kavitharaju  Will it be fine not using `fastapi[all]`? we had `pydantic 1.10` with`fastapi 0.95` . Once we upgrade to `fastapi 0.101.0` by keeping pydantic `v1` we need to downgrade pydantic(than existing one) to avoid dependency conflicts.
This can be avoided by using only `fastapi` and separately mentioning required packages.

Joel-Joseph-George: if there is anymore packages we need , please mention by.. these were the ones i could find by.

Joel-Joseph-George:
**Noticed issue while running uvicorn app..**
Screenshot from 2023-08-22 11-14-53

this will be the main.py
image

Issue #674, opened 2023-08-21
Update Terminology: Change /v2/bibles -> /v2/resources/bibles


shimilgithub:
Rename bible related end points from `/v2/bibles...` to `/v2/resources/bibles...`
[Discussions](https://github.com/Bridgeconn/vachan-api/discussions/575)

PR #675, opened 2023-08-21
renaming endpoint : bibles

shimilgithub:
- #674


kavitharaju (commented):
there was a place where we were deleteing only .first(), where it should have been .all(). Is that change included?

shimilgithub:
> there was a place where we were deleteing only .first(), where it should have been .all(). Is that change included?

No. Separate issue created for that

Issue #688, opened 2023-08-31
Translation Projects shows same createTime


shimilgithub:
Translation Projects created at different times shows same `createTime` value
![MicrosoftTeams-image (2)](https://github.com/Bridgeconn/vachan-api/assets/54615511/641c97a8-9ec4-4a64-8b64-f4272671d9d1)

PR #689, opened 2023-08-31
translation projects shows same createTime

shimilgithub:
- #688

### DIFF
--- a/app/test/test_agmt_projects.py
+++ b/app/test/test_agmt_projects.py
@@ -1,4 +1,5 @@
 '''Test cases for Agmt projects related APIs'''
+import time
 from . import client
 from . import assert_input_validation_error, assert_not_available_content
 from . import check_default_get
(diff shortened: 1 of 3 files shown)


app/test/test_agmt_projects.py
```diff
@@ -906,6 +907,26 @@ def test_create_n_update_times():
...
+    assert response1.status_code == 201
+    assert response1.json()['message'] == "Project created successfully"
+    new_project2 = response1.json()['data']
+    assert_positive_get(new_project2)
+    print("///post resp1:",new_project2)
+    create_time_2 = new_project2['createTime']
+    update_time_2 = new_project2['updateTime']
+    assert create_time_2 == update_time_2
+
+    assert not new_project['createTime'] == create_time_2
```
kavitharaju, line 927: Good to add a comment with the issue number or link here. Otherwise, someone reading this later will find it confusing why there is a test case like this

kavitharaju (approved):
LGTM, other than the minor comment suggestion

Issue #429, opened 2022-10-03
Kratos Version change


sijumoncy:
upgrade kratos to new version v0.10.1

kavitharaju:
Need to look into the issue happened upon updating the Kratos version on staging server.

First the vachan-api contaier was not starting successfully, upon code update after auto deployment.
Upon restarting the Kratos DB(that is run separately, with database.yml) miragte started to run for a long time, as if creating new DB, and ended in error saying the roles are not existing in DB

![MicrosoftTeams-image (4)](https://user-images.githubusercontent.com/18255929/193742209-40742c17-2c8c-4aa7-bd25-4b439ca8aba3.png)

Then, after the code changes were reverted, it was found that, the DB got created fresh(with older Kratos version), thus loosing all the user data we had.

Need to look into this matter, by testing it locally and ensuring that 
- Kratos version update does not make old DB incompatible and also 
- will not re-create the DB , if an older one exists.

kavitharaju:
https://github.com/ory/kratos/releases/tag/v1.0.0

PR #706, opened 2023-09-14
Upgrade Kratos to v1.0.0

shimilgithub:
- #429 
- Only `lastname` is kept as optional for registration

### DIFF
--- a/app/auth/authentication.py
+++ b/app/auth/authentication.py
@@ -495,7 +495,8 @@ def register_check_success(reg_response):
         "registered_details":{
             "id":reg_response["identity"]["id"],
             "email":reg_response["identity"]["traits"]["email"],
-            "Name":str(name_path["first"]) + " " + last_name,
+            "Name": str(name_path["first"]) + (" " + last_name if last_name else ""),
+
             "Permissions": reg_response["identity"]["traits"]["userrole"]
         },
         "token":reg_response["session_token"]
--- a/app/auth/authentication.py
+++ b/app/auth/authentication.py
@@ -386,11 +386,14 @@ def get_users_kratos_filter(base_url,name,roles,limit,skip):#pylint: disable=too
                 "userId":data["id"],
                 "name":data["traits"]["name"]
             }
-            kratos_user["name"]["fullname"] = data["traits"]["name"]["first"].capitalize() \
-                + " "+ data["traits"]["name"]["last"].capitalize()
+
+            first_name = data["traits"]["name"]["first"].capitalize()
+            last_name = data["traits"]["name"].get("last", "")
+            last_name = last_name.capitalize() if last_name else last_name
+            kratos_user["name"]["fullname"] = first_name+ " " + last_name
+
             if not name is None:
                 if name.lower() == kratos_user["name"]["fullname"].lower() or\
-                    name.lower() == kratos_user["name"]["last"].lower() or\
                         name.lower() == kratos_user["name"]["first"].lower():
                     name_status = True
                 else:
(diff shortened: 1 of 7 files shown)


app/auth/authentication.py
```diff
@@ -486,12 +489,13 @@ def update_kratos_user(rec_user_id,data):
...
     """register reqirement success"""
     name_path = reg_response["identity"]["traits"]["name"]
+    last_name = name_path.get("last", "")
     data={
         "message":"Registration Successfull",
         "registered_details":{
             "id":reg_response["identity"]["id"],
             "email":reg_response["identity"]["traits"]["email"],
-            "Name":str(name_path["first"]) + " " + str(name_path["last"]),
+            "Name":str(name_path["first"]) + " " + last_name,
```
kavitharaju: would this add an unnecessary space after first name if lastname is ''?

docker/Kratos_config/email-password/kratos.yml
```diff
@@ -12,7 +12,7 @@ serve:
 
 selfservice:
   default_browser_return_url: http://127.0.0.1:4455/
-  whitelisted_return_urls:
+  allowed_return_urls:
     - http://127.0.0.1:4455
```
kavitharaju: Will hardcoding this to 127.0.0.1 casue issues in deployments and docker?
Can we use the domain ENV variable in this file? Please check if there are other places where we use ENV variable in this file...

docker/docker-compose.yml
```diff
@@ -91,7 +91,7 @@ services:
      - 5432
     ports:
      # HOST:CONTAINER
-     - "5433:5432"
+     - "5439:5432"
```
kavitharaju: You might have needed this change for your local testing because that port is already in use. But I am not sure if we need to change that in the common repo. this will have impact in the the connection configs we use on PGAdmin and all to connect to the server DBs

docker/Kratos_config/email-password/kratos.yml
shimilgithub: @kavitharaju This file doesn't use env variables anywhere. But this IP is hard coded in many places.

kavitharaju: Ah, its not using ENV vars. But uses the docker-service name instead, like this...
https://github.com/Bridgeconn/vachan-api/blob/7e1174af80a53dee065514f386780c36fee86b2f/docker/Kratos_config/email-password/kratos.yml#L7

Please see if this is what we need

app/auth/authentication.py
```diff
@@ -386,11 +386,14 @@ def get_users_kratos_filter(base_url,name,roles,limit,skip):#pylint: disable=too
                 "userId":data["id"],
                 "name":data["traits"]["name"]
             }
-            kratos_user["name"]["fullname"] = data["traits"]["name"]["first"].capitalize() \
-                + " "+ data["traits"]["name"]["last"].capitalize()
+
+            first_name = data["traits"]["name"]["first"].capitalize()
+            last_name = data["traits"]["name"].get("last", "")
+            last_name = last_name.capitalize() if last_name else last_name
+            kratos_user["name"]["fullname"] = first_name+ " " + last_name
```
sijumoncy: does this make an empty space at the end of fullname if lastname is not there ?

sijumoncy:
@shimilgithub  Can you check and ensure the mail is getting to user on registration properly with version upgrade. Just to ensure no changes happened in the URL for Mailing

you can check it in local mail server on development.

sijumoncy (approved):

kavitharaju (approved):

Issue #599, opened 2023-06-23
add chapter range in commentary api


AnuMonachan:
Give option to have `chapterStart` and `chapterEnd` to store commentary text that spans across multiple chapters instead of just `chapter`.

e.g:-    `Gen. 11:27–50:26 Patriarchal History. The story now moves from the general survey of humanity to the specific family from which Israel comes.` This commentary text spans across chapters 11 - 50 in [esv.org](https://www.esv.org/Genesis+11/)
* Commentary GET calls for all chapters from Gen 11 to Gen 50 should fetch this text.

kavitharaju:
@AnuMonachan  thank you for this suggestion

As we are handling this case of verse ranges spanning across chapters, we would also like to asses another scenario and see how frequent and important that is in data.
Can you help us check the data for **discontinuous list of verses** or verse-ranges being associated to a commentary?
eg.: One commenatry snippet having its reference like this `MAT 1:3, MAT 5:6, MAT 10:20` or `MAT 1:3-10, MAT 5:6, MAT 10:20-15:5`

AnuMonachan:
I have not come across discontinuous list of verses like the above-mentioned one while uploading esv or bbc commentaries, so far it has been continuous ranges like 1:2-10 or 1:2-2:5

kavitharaju:
Okay, thanks for letting us know :+1:

RevantCI:
Hi Kavitha, I just had a thought if you wanted to investigate more, The raw data commentary may have  hints about the discontinuous verse list data you inquired about. They could be in the form of foot notes or cross references.
I think the content team gets USFX files with USFM tags sometimes, which are not included in the cleaned csv files they give us the VO/VA team.
If you wanted to see the raw USFX data/ process you could check with Samuel or Vishnu.

shimilgithub:
> Hi Kavitha, I just had a thought if you wanted to investigate more, The raw data commentary may have hints about the discontinuous verse list data you inquired about. They could be in the form of foot notes or cross references. I think the content team gets USFX files with USFM tags sometimes, which are not included in the cleaned csv files they give us the VO/VA team. If you wanted to see the raw USFX data/ process you could check with Samuel or Vishnu.

@samuelanthrayose @svishnu06  Do you have anything to add here?

PR #715, opened 2023-09-26
Changing reference structure and related tests in commentaries

shimilgithub:
- #599 
- `reference` structure of commentaries has been changed
- can handle discontinuous verses
- added filtering with `search word`
- PUT API input schema changed to edit commentary by fetching unique `commentaryId` instead of `reference`


app/crud/contents_crud.py
```diff
@@ -74,48 +79,48 @@ def upload_commentaries(db_: Session, resource_name, commentaries, job_id, user_
...
+                else:
+                    #setting verseEnd to 999 if its not present
+                    ref_end  = utils.create_decimal_ref_id(db_,ref['bookEnd'],ref['chapterEnd'],999)
+                    ref['verseEnd'] = 999
+            else:
+                ref_end = None
+        else:
+            ref = None
+            ref_end = None
+            ref_start = None
```
kavitharaju: A few questions:
1. What if `ref['chapter']` is also `None` along with `ref['verseNumber']`? Shouldn't chatper also be set to 0 indicating the reference applies to full book?
2. Similary `ref['chapterEnd']` could also be `None`. For example user add book intro for Book of Matthew and could give just the book value as rest is intutive
3. Why do we set values to None and not raise error in case of not being able to compose a reference?
4. Isn't same or similar logic present in other places like add and edit of parascripturals, sign videos, audio bibles etc. Can't we create a single function to compose reference values from user input and ensure consistent performance across different resource types?

app/crud/contents_crud.py
```diff
@@ -74,48 +79,48 @@ def upload_commentaries(db_: Session, resource_name, commentaries, job_id, user_
...
+        query = db_.query(model_cls)
+        for row in query.all():
+            exist_check = query.filter(model_cls.refStart == ref_start).first()
             if exist_check:
                 update_args["output"]= {
                 "message": 'Already exist commentary with same values for reference range',
```
kavitharaju: Shouldn't we check for end also before raising error? I guess we will also need to check with type once we have that feature built in as reference could repeat across different types.

app/db_models.py
```diff
@@ -121,34 +121,38 @@ class BibleBook(Base): # pylint: disable=too-few-public-methods
...
+
+    _reference = Column('reference', JSONB)
+
     @hybrid_property
     def ref_string(self):
         '''To compose surrogate id'''
-        return f'{self.book.bookCode} {self.chapter}:{self.verseStart}-{self.verseEnd}'
+        return f'{self.reference["book"]} {self.reference["chapter"]}:\
+            {self.reference["verseNumber"]}-{self.reference["verseEnd"]}'
+
```
kavitharaju, line 145: Oh is this the format we are using... I thought we have book, chapter, verse in refStart and book,chapter, verse in refEnd... allowing us to do cross-chapter and cross-book references?!!!

app/schema/schema_content.py
```diff
@@ -274,48 +241,17 @@ class Config:
...
         '''display example value in API documentation'''
         schema_extra = {
             "example": {
-                "bookCode": "1ki",
-                "chapter": 10,
-                "verseStart": 1,
-                "verseEnd": 7,
+                "commentaryId":100000,
+                "reference": {"book":"MRK", "chapter":15, "verseNumber":10,
+                               "bookEnd":"JHN", "chapterEnd":1, "verseEnd":9 },
```
kavitharaju, line 254: See here there is that option I was talking about above... But why is that function different?

app/db_models.py
shimilgithub: did you meant to add `bookEnd` and `chapterEnd` also here?  But this part is used to [extract text](https://github.com/Bridgeconn/vachan-api/blob/version-2/app/crud/contents_crud.py#L1335C35-L1335C35).

Cross chapter reference is already [handled ](https://github.com/Bridgeconn/vachan-api/blob/55154bfa5e61dd258b03e895eddb69646402fcb5/app/test/test_commentaries.py#L303)with `refStart` and `refEnd`

app/crud/contents_crud.py
shimilgithub: does reference range can be duplicated in commentary like same reference range with different commentary. Only `commentaryId` is set as unique in SQLAlchemy Model

AthulyaMS (dismissed):

app/crud/contents_crud.py
kavitharaju: One way to do it could be to have unique constraint over refStart, refEnd and type (that is going to come).
But I am also not opposed to the idea of not keeping such a unique constraint at all

app/db_models.py
kavitharaju: Oh , I didn't realize it was part of extract text....
but now looking at extract text, I think its time for us to refactor that based on new references we keep as well as new resource types like parascriptural and all... But we can track and take it up as a different task. For now, ensure older features are working :+1:

app/crud/contents_crud.py
shimilgithub: 1.&2 handled in [utils crud](https://github.com/Bridgeconn/vachan-api/blob/version-2/app/crud/utils.py#L238), made some changes and added similar data in test(test_post_default)
3. `reference` is an optional field in `POST` operation. SO these fields are set to `None` in case it is absent.

kavitharaju (approved):
I hope the raised issues are handled and tested to be working. (not doing a line by line check again)

Issue #635, opened 2023-07-14
Option to add commentary book intro to the get commentary API call


RevantCI:
As per the new [feature request](https://github.com/Bridgeconn/VachanOnline-v2/issues/212) in vachan online we are showing the book intro along with each chapter of the commentary. This helps to understand the context of the book.
To do that currently we are fetching the book intro by making a call to chapter 1 of the book ( which send the book intro) along with the chapter being read.
To avoid making 2 API calls each time can the book intro be sent along with each chapter of the commentary in the [get](https://github.com/Bridgeconn/VachanOnline-v2/issues/212) commentary call
If the use case may not be used by everyone a flag can be used to indicate if the book intro should be sent or not in the response.

kavitharaju:
The usecase definitely makes sense. I wonder if we can add provision to accept multiple verse/verse-ranges in a single API call?

At the DB level it would be two where clauses combined by `or`. Something like this 
```
start_refid <= ref_id1 <= end_refid or start_refid <= ref_id2 <= end_refid
```
At the IO, can we have a query param that expects a list of verse/verse-ranges? 
If the above is possible, the usage could be something like this 
```
const res = await axios.get(
                            '.../v2/commentaries/{source_name}', { 
                            params: { reference: [
                                     {book: VAL chapterStart:0},  // for book intro
                                     {book: VAL, chapterStart: 5, verseStart:0, verseEnd:150}   //for the specific chapter's content
] } });
```  

This would a major refactor on the back end. But these are things we did not re-consider while copying directly from V1. Its worth discussing this and deciding a general solution as per our use-cases, present and future.

Worth considering 
- this requirement  #599  and 
- the verse reference implementation we are planning for parascripturals https://github.com/Bridgeconn/vachan-api/pull/632#discussion_r1262626277

Current commentary table in DB
![image](https://github.com/Bridgeconn/vachan-api/assets/18255929/22b3ae0b-1b77-4360-a748-199a59b375b3)

Change proposed by me
- remove of book_id, chapter, verse_start, verse_end
- newly add ref_start and ref_end as bbcccvvv integer fields.
To handle cross-chapter references representation and this requirement of discontinuous reference query, efficiently using integer operations at SQL level

joelthe1:
[...]

PR #731, opened 2023-10-05
Handling special cases in commentary with new field

shimilgithub:
- #635 
- **Has database change**
- new field `sectionType` introduced which can be of type `commentary-text`,`book-introduction`,`epilogue` etc.

### DIFF
--- a/app/routers/content_apis.py
+++ b/app/routers/content_apis.py
@@ -663,7 +663,7 @@ async def get_bible_verse(request: Request,
 @get_auth_access_check_decorator
 async def get_commentary(request: Request,
     resource_name: schemas.TableNamePattern=Path(..., examples="en_BBC_1_commentary"),
-    section_type:schema_content.SectionTypes=Query(schema_content.SectionTypes.COMMENTARY),
+    section_type: List[str] = Query(None, examples=["commentary-text", "book-introduction"]),
     reference: str = Query(None,
     examples='{"book": "mat", "chapter": 1, "verseNumber": 6}'),
     search_word: str=Query(None, examples="customary") ,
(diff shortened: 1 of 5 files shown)


app/routers/content_apis.py
```diff
@@ -663,6 +663,7 @@ async def get_bible_verse(request: Request,
 @get_auth_access_check_decorator
 async def get_commentary(request: Request,
     resource_name: schemas.TableNamePattern=Path(..., examples="en_BBC_1_commentary"),
+    section_type:schema_content.SectionTypes=Query(schema_content.SectionTypes.COMMENTARY),
```
kavitharaju: If we set a default value to that would lead the user to never get all the different types present in one request. I recommend making the default to `None`, retrieving all types if user doesn't specify one.

Also I think making this an array field, allowing multiple values to be passed would make a big improvement in the API. For example, a request to fetch, the "intro" and "normal commentary" for a chapter in one go would be useful for the clients

app/schema/schema_content.py
```diff
@@ -8,6 +8,14 @@
 from crud import utils
 
 #pylint: disable=too-few-public-methods
+
+class SectionTypes(str, Enum):
+    '''To specify resource access permisions'''
+    COMMENTARY = "commentary-text"
+    BOOKINTRO = "book-introduction"
+    EPILOGUE = "epilogue"
+    TOPICOUTLINE = "topic-outline"
```
kavitharaju: I dont think we should define these possible types. That will only limit the users from using all diverse types out there... For instance, this list does not cover "chapter-intros". Lets not try to know what all are out there, but just let users provide what ever value they want to. 
Only the default value, the one we use if no value is provided, need to be defined by us. It could be the above "commentary-text"

AthulyaMS (approved):

Issue #732, opened 2023-10-09
 Vachan Engine ER Diagram


shimilgithub:
Needed an  ER diagram illustrating the core data relationships within the Vachan Engine, facilitating easy understanding and efficient database management.

kavitharaju:
Could refer this for inspiration https://github.com/Bridgeconn/vachan-api/discussions/575#discussioncomment-6238855

PR #738, opened 2023-10-17
Entity Relationship Diagram of VE

shimilgithub:
- #732


kavitharaju (commented):
1. Translation module's tables could be excluded if not needed. Would simplify the diagram. It could be made separately, if needed.
2. Opening this as an image in another app like Powerpoint or flow-chart makers will allow you to add understandable names to the dynamic tables

shimilgithub:
> 1. Translation module's tables could be excluded if not needed. Would simplify the diagram. It could be made separately, if needed.
> 2. Opening this as an image in another app like Powerpoint or flow-chart makers will allow you to add understandable names to the dynamic tables

@kavitharaju  Can it be kept as pdf for more clarity?

kavitharaju:
Yes, once you have a good diagram created in image formats, you can then convert to PDF and keep it in the repo :+1:

AthulyaMS (approved):

PR #745, opened 2023-11-21
Add SSL setup script for Dockerized Nginx with Let's Encrypt

udkumar:
This commit introduces a new shell script for setting up and managing a Dockerised Nginx environment. Key features include:

- Stopping/starting Nginx container
- Freeing up port 80 on the host
- Installing Nginx and Certbot if not present
- Generating Let's Encrypt SSL certificates
- Setting up automatic SSL certificate renewal
- Copying SSL certificates to a specified Docker volume

The script is intended to streamline the deployment and maintenance of secure Nginx setups in Docker environments.

### DIFF
--- a/docker/auto_ssl_script.sh
+++ b/docker/auto_ssl_script.sh
@@ -4,99 +4,70 @@
 # Developer Details
 # ------------------------------------------------------------------------------
 # Name: Uday Kumar
-# Contact: [uday.kumar@bridgeconn.com](uday.kumar@bridgeconn.com)
+# Contact: uday.kumar@bridgeconn.com
 # Date: 21 Nov 2023
-# Description: This script sets up and manages a Dockerized Nginx environment 
-# with Let's Encrypt SSL, including automatic certificate renewal.
+# Description: This script sets up and manages Let's Encrypt SSL, including automatic certificate renewal.
 # ------------------------------------------------------------------------------
 
-# Environment variables
-NGINX_CONTAINER_NAME="${NGINX_CONTAINER_NAME}"
-DOMAIN_NAME="${VACHAN_DOMAIN}"
-EMAIL="${EMAIL}"
 
-# Can change as per host letsencrypt path
-CERT_PATH="/etc/letsencrypt/."
-DOCKER_VOLUME_PATH="./certbot/conf/"
+# Load environment variables
+if [ -f prod.env ]; then
+    source prod.env
+else
+    echo "prod.env file not found"
+    exit 1
+fi
 
-# Stop Nginx container
-stop_nginx_container() {
-    echo "Stopping Nginx container: $1..."
-    docker stop "$1" || { echo "Failed to stop Nginx container"; exit 1; }
-}
+# Configuration variables
+NGINX_CONTAINER_NAME="docker-web-server-with-cert-1"
+SSL_CERTS_DIR="/certbot/conf"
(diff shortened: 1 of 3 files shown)


kavitharaju (changes requested):
In our existing setup we have nginx and certbot as docker services and we avoid installing them directly on the host machine. Please update this script to work with the docker services. 

How/when to run this shell script? Please document necessary details.

udkumar:
I'm currently utilising Docker services exclusively and not using Nginx on the host. 

I've automated the manual processes with a single shell script, which accomplishes the following tasks:

1. Executes our Docker Compose based on the provided steps.
2. Stops the Nginx container.
3. Installs Nginx on the host and sets up SSL for a domain using Let's Encrypt with the aid of Certbot.
4. Transfers the SSL certificate into a volume designated for the Nginx container.
5. Then, it stops the host's Nginx service, which can also be removed if necessary.
6. Finally, it restarts the Nginx container that was initially stopped in step 2

udkumar:
@kavitharaju 
I have updated with the following changes:
 

- No manual intervention is required.
- The existing docker compose command (added in readme) will handle the SSL certificate and renew it automatically every 80 days.
- I have modified the shell script to include only the relevant steps.

docker/auto_ssl_script.sh
```diff
@@ -0,0 +1,73 @@
...
+# Load environment variables
+if [ -f prod.env ]; then
+    source prod.env
+else
+    echo "prod.env file not found"
+    exit 1
+fi
+
+# Configuration variables
+NGINX_CONTAINER_NAME="docker-web-server-with-cert-1"
```
joelthe1, line 22: Is there a way to not hardcode this and get this from the env?

docker/auto_ssl_script.sh
```diff
@@ -0,0 +1,73 @@
...
+SSL_CERTS_DIR="/certbot/conf"
+WEBROOT_DIR="/certbot/www"
+
+# Generate SSL certificates
+generate_certificates() {
+    echo "Generating SSL certificates for $DOMAIN..."
+
+    docker run --rm -it \
```
joelthe1, line 30: So does this run docker within docker?

docker/docker-compose.yml
```diff
@@ -267,25 +267,29 @@ services:
...
-   - ./certbot/www/:/var/www/certbot/:rw
-   - ./certbot/conf/:/etc/letsencrypt/:rw
+    - ./certbot/conf:/etc/letsencrypt
+    - ./auto_ssl_script.sh:/auto_ssl_script.sh
   profiles:
    - deployment
+  entrypoint: /bin/sh -c '
+    while :; do
+      /auto_ssl_script.sh;
+      sleep 80d;
```
joelthe1, line 289: Is there a way to use the ofelia-scheduler which is already used for scheduled backups for this task?

docker/auto_ssl_script.sh
udkumar: This we can take from env for sure

docker/auto_ssl_script.sh
udkumar: Yes, I also checked that the current approach isn't ideal. If it's not suitable, I'm working on creating a single Docker solution, though it's taking some time.

After your comment, I tried but I am getting the same issue as certificate not creating.

Issue #836, opened 2024-06-24
Utilize newly allocated space in production(backup)


AthulyaMS:
- [x] Understanding about backups in production(3 hr)
- [x] Utilize newly allocated space in production(backup)(4 hr)
- [x] PR Review-1 hr(shimil)
- [x] Review fix -1hr

AthulyaMS:
- [x] Check whether db error occuring in 1 week -0.5 hr
- [x] Learn deployment workflow -6.5hr
- [x] Make different docker compose for staging and production -3hr
- [x] PR Review -1hr(shimil)
- [x] Review fix -2hr

PR #850, opened 2024-07-18
Seperate  docker compose file in staging and production for backup

AthulyaMS:
- #836 

-   Separate Docker Compose files for staging and production. In the staging Docker Compose file, there is no Ofelia scheduler container, while in the production Docker Compose file, the Ofelia scheduler is included for backup


docker/docker-compose-production.yml
shimilgithub, line 1: why this file is empty?

.github/workflows/run-test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Start dependancies
         working-directory: ./docker
-        run: docker compose up -d
+        run: docker compose -f docker-compose-staging.yml up -d
```
shimilgithub, line 21: does this needs to happen only for staging?

.github/workflows/run-test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
...
       - name: Run tests
         working-directory: ./docker
-        run: docker compose run vachan-api-test
\ No newline at end of file
+        run: docker compose -f docker-compose-staging.yml run vachan-api-test
```
shimilgithub, line 24: here also

docker/docker-compose-production.yml
AthulyaMS: I dont know how that file become empty .I updated it again

.github/workflows/run-test.yml
AthulyaMS: @shimilgithub Actually do we need this built-n-test in staging and production ?
Since we have both compose file now ,inorder to run the built-n-test ,we need to specify one thats why I added staging here .I think this file is using at the stage of local to github .

shimilgithub: As per this , during release also staging docker compose will run

shimilgithub (approved):

PR #854, opened 2024-07-23
Documentation updates after docker compose changes

shimilgithub:
-  #836 
-  Includes changes in documentation after separating it into `docker-compose-staging` and `docker-compose-production`

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,14 +18,14 @@ Implemented Using
 
 ```cd vachan-api/docker```
 
-```docker compose --profile local-run up```
+``` docker compose -f docker-compose-staging.yml --profile local-run up```
 or
-```sudo -E docker compose --profile local-run up```
+```sudo -E  docker compose -f docker-compose-staging.yml --profile local-run up```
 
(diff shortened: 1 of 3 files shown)


AthulyaMS:
I didnt see a comand to down docker compose ,do we need to add that .

docker compose -f docker-compose-production.yml --profile deployment --env-file prod.env down

The above command was worked in production to down compose .If this is needed ,update in staging also

shimilgithub:
> I didnt see a comand to down docker compose ,do we need to add that .
> 
> docker compose -f docker-compose-production.yml --profile deployment --env-file prod.env down
> 
> The above command was worked in production to down compose .If this is needed ,update in staging also

can you point to the right place in PR?

AthulyaMS:
Actually, in the documentation, I can't see how to down the Docker Compose. I was asking for a suggestion: should we add a command to bring down the Compose file?

shimilgithub:
> docker compose -f docker-compose-production.yml --profile deployment --env-file prod.env down

updated

AthulyaMS (approved):

PR #890, opened 2025-02-11
integration of Anything llm app

Jayasankar-kk:
- added anythingllm container in staging and production compose files with necessary env variables
- nginx default.conf and app.conf.template modified to accomodate anythingllm app using a new subdomain anythingllm.vachanengine.org


kavitharaju (commented):
Aren't you passing the anything llm subdomain to the `
 web-server-with-cert` container as env_var, `VACHAN_LLM_DOMAIN` ?

PR #911, opened 2026-01-09
Adding New version for bible-glyph and add in same docker network

AthulyaMS:
- Added new dockerimage
- Made the bible-glyph container under VE network
- Change the port in nginx config file to 8000

### DIFF
--- a/docker/docker-compose-production.yml
+++ b/docker/docker-compose-production.yml
@@ -654,7 +654,7 @@ services:
      - VE-network
 
  bible-glyph:
-    image: kavitha3797/bible-glyph:latest
+    image: kavitha3797/bible-glyph:1.0.0-alpha.3
     ports:
       - "8011:8000"
     environment:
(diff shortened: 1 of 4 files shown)


docker/docker-compose-production.yml
```diff
@@ -654,7 +654,7 @@ services:
      - VE-network
 
  bible-glyph:
-    image: kavitha3797/bible-glyph:latest
+    image: kavitha3797/bible-glyph:1.0.0-alpha.3
```
kavitharaju: Its 1.0.0-alpha.4

kavitharaju (changes requested):

docker/docker-compose-production.yml
AthulyaMS: changed in next commit

kavitharaju (approved):
LGTM

